PR #96, opened 2015-12-28
BROWSER-95: Refactor cliqz.cfg

maxim-cliqz:


mozilla-release/browser/confvars.sh
```diff
@@ -35,8 +35,8 @@ MOZ_SERVICES_COMMON=1
 MOZ_SERVICES_CRYPTO=1
 MOZ_SERVICES_HEALTHREPORT=1
 MOZ_SERVICES_METRICS=1
-MOZ_SERVICES_SYNC=1
```
chrmod: maybe we can find a better way to alter those vars. One that will not require changes in source files

chrmod:
retest this please

mozilla-release/browser/confvars.sh
maxim-cliqz: BTW, we may also want to disable mozilla health report service (see line 36). @luciancor ?

alver-cliqz: Better to redirect this to our URL and put a dummy handler for it. Never knows when it could be useful for us

PR #103, opened 2015-12-30
Build: clean old *.pyc

chrmod:
@maxim-cliqz please verify if that solves our problem

### DIFF
--- a/magic_build_and_package.sh
+++ b/magic_build_and_package.sh
@@ -61,6 +61,11 @@ export MOZ_AUTOMATION_UPLOAD=1
 
 ./mach clobber
 
+# Clean old compiled python files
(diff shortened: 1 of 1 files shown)


magic_build_and_package.sh
```diff
@@ -61,6 +61,11 @@ export MOZ_AUTOMATION_UPLOAD=1
 
 ./mach clobber
 
+# Clean old compiled python files
+# Can be removed after merge with Firefox 43
+# Fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=1184696
+find -name \*.pyc -delete
```
alver-cliqz: find . -name \*.pyc -delete

PR #105, opened 2016-01-04
Build: updating CLIQZ upload scripts

chrmod:
Affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1198179


alver-cliqz:
Some copy-paste from original UploadFiles functions, must be changed in further FF versions.

PR #139, opened 2016-01-18
enforce extension signature

luciancor:
reverting the change from @alver-cliqz 
cc @chrmod


luciancor:
merging this to start the build

PR #141, opened 2016-01-27
Create branding system

alver-cliqz:
Script for inject tag information into installer (will be used on builder).
Some code cleaning and formatting.


maxim-cliqz:
Looks OK

PR #149, opened 2016-01-29
Removing english dictionary from german builds

chrmod:


luciancor:
closing this - we need to find a more sane solution

PR #153, opened 2016-02-01
BROWSER-108: fix crash on Win7

alver-cliqz:
Incorrect use of WriteFile WinAPI function


mozilla-release/other-licenses/nsis/Contrib/CliqzHelper/dllmain.cpp
```diff
@@ -36,6 +36,7 @@ extern "C" void __declspec(dllexport) setBrand(HWND hwndParent,
                                                TCHAR *variables,
                                                stack_t **stacktop) {
 #ifdef _DEBUG
+  __debugbreak();
```
chrmod: should this one be commited?

alver-cliqz: no, it used only in Debug build

PR #154, opened 2016-02-01
Branding: images update

chrmod:
New set branding images that match official mozilla branding format.
Removing not used files as well.

Hope to solve pixelated icon problem on Window - https://cliqztix.atlassian.net/browse/DB-116


chrmod:
This one should not be merged, sorry for not marking it properly. Will fix it and let know when ready.

PR #162, opened 2016-02-04
DB-98: fix bug in FF configuration script with locale processing

alver-cliqz:
It's look like possible to use two different ways to set up a locale for building - MOZ_UI_LOCALE environment variable or with mozconfig file with --enable-ui-locale option. In real - MOZ_UI_LOCALE always set up in configure.in, so it's only one way to change it - with mozconfig file. Which looks wrong. Here we make changes in configure.in script so now it's possible to set up language correctly in both different way.

ps. looks like js/src/configure.in have similar problem, but it don't used in our build now
pps. may be we can push this commit to original FF


maxim-cliqz:
looks good to me

PR #163, opened 2016-02-04
DB-311: Allow both session restore and fresh tab

maxim-cliqz:


mozilla-release/browser/app/profile/firefox.js
```diff
@@ -1272,7 +1272,8 @@ pref("services.sync.prefs.sync.browser.safebrowsing.malware.enabled", true);
 pref("services.sync.prefs.sync.browser.search.update", true);
 pref("services.sync.prefs.sync.browser.sessionstore.restore_on_demand", true);
 pref("services.sync.prefs.sync.browser.startup.homepage", true);
-pref("services.sync.prefs.sync.browser.startup.page", true);
+pref("services.sync.prefs.sync.browser.startup.restoreTabs", true);
+pref("services.sync.prefs.sync.browser.startup.addFreshTab", true);
```
maxim-cliqz: Not sure, what exactly this affects. Just did by example

mozilla-release/browser/components/sessionstore/nsSessionStartup.js
```diff
@@ -314,12 +314,15 @@ SessionStartup.prototype = {
    * @returns bool
    */
   get willOverrideHomepage() {
+    // Cliqz allows both restoring previous session and opening a home page.
+/*  // TODO: Replace with #if 0
```
maxim-cliqz: I'd like to fix this before merging, but it's not critical

mozilla-release/toolkit/components/telemetry/TelemetryEnvironment.jsm
```diff
@@ -108,7 +108,8 @@ const DEFAULT_ENVIRONMENT_PREFS = new Map([
   ["browser.shell.checkDefaultBrowser", {what: RECORD_PREF_VALUE}],
   ["browser.search.suggest.enabled", {what: RECORD_PREF_VALUE}],
   ["browser.startup.homepage", {what: RECORD_PREF_STATE}],
-  ["browser.startup.page", {what: RECORD_PREF_VALUE}],
+  ["browser.startup.restoreTabs", {what: RECORD_PREF_VALUE}],
+  ["browser.startup.addFreshTab", {what: RECORD_PREF_VALUE}],
```
maxim-cliqz: Again, not sure, what is the exact effect of this

mozilla-release/browser/app/profile/firefox.js
luciancor: @maxim-cliqz we use that pref in the extension and it seems a bit weird to add 2 new prefs for what we want to do. What about adding a new value in browser.startup.page? eg: 4 :) here is the list with all the values - http://kb.mozillazine.org/Browser.startup.page

mozilla-release/toolkit/components/telemetry/TelemetryEnvironment.jsm
luciancor: @maxim-cliqz building on the previous comment - using a different value will allow us to keep this file as it was with no changes

mozilla-release/browser/app/profile/firefox.js
maxim-cliqz: Technically it's possible, but will add more complexity to preference value handling, because we have 2 separate controls now. So, 2 separate values actually just reflect that those are independent preferences now.

mozilla-release/toolkit/components/telemetry/TelemetryEnvironment.jsm
maxim-cliqz: This is a trivial conflict to fix, if it will ever appear.
Also, I don't think that trying not to touch original Firefox files is of highest priority.

mozilla-release/browser/app/profile/firefox.js
luciancor: cant we add something like:
4 - FreshTab
5 - restoreTabs and freshtab
technically restoreTabs is option 3 from http://kb.mozillazine.org/Browser.startup.page right?

if we choose to add different preferences we would simply move the complexity from the browser preferences to the extension

PR #168, opened 2016-02-09
DB-354: MacOS bundle tagging

maxim-cliqz:
Right now script puts the tag into `distribution.js` pref file and into `Info.plist` meta-info file inside the bundle.
First one is used by our extension, AFAIK.
The second one may be used by native code in case we need to know the current running tag, not the one who was installed first.
Problem is: `distribution.js` is also replaced if app is reinstalled. In order to preserve it, we must copy that data either to the profile, or somewhere in the user home directory.
Ideas?

### DIFF
--- a/cliqz-helpers/dmg.json
+++ b/cliqz-helpers/dmg.json
@@ -1,0 +1,20 @@
+{
+  "title": "",
+  "background": "../mozilla-release/browser/branding/cliqz/background.png",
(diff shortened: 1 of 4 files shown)


maxim-cliqz:
cc @alver-cliqz , @chrmod

maxim-cliqz:
Will go in another repo

PR #171, opened 2016-02-10
DB-365: Homepage preference is reset after restart

maxim-cliqz:
This preference default value is already set in `browser/app/profile/firefox.js`, value is taken from browserconfig.properties.


mozilla-release/browser/branding/cliqz/locales/browserconfig.properties
```diff
@@ -3,4 +3,4 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # Do NOT localize or otherwise change these values
-browser.startup.homepage=about:home
+browser.startup.homepage=about:cliqz
```
maxim-cliqz: Not strictly necessary, as this url is overridden in `cliqz.cfg`.
Just for the sanity.

maxim-cliqz:
retest this please

PR #172, opened 2016-02-10
DB-372: improve digital signing script on Windows

alver-cliqz:
Now works with all *.exe and *.dll files in installer. Also check does file have been signed before or not. And fix an error handling (work with ERRORLEVEL in windows batch script is tricky, https://support.microsoft.com/en-us/kb/39585)


luciancor:
@chrmod can you please check this PR?

PR #173, opened 2016-02-11
Release

luciancor:
@alver-cliqz port fixes from release back to master


alver-cliqz:
@luciancor as i see, my fixes for branding installer already in master. Only your changes with new extension version could not be applied for master, because @maxim-cliqz change build script and they are placed in another place. I closed this PR, let's create separate one with extension update only.

PR #187, opened 2016-02-22
Update sign_mac.sh

chrmod:
There is a problem in signing process. Somehow the DMG that we generate hosts unsigned app. 
CLIQZ.app in pkg folder is signed properly.

### DIFF
--- a/sign_mac.sh
+++ b/sign_mac.sh
@@ -14,11 +14,16 @@ do
   for app in $PKG_DIR/*.app
   do
     codesign -s $CQZ_CERT_NAME --force --deep $app
+    if [ $? -ne 0 ]; then
+      echo "Something went wrong during signing"
+      exit 1
+    fi
   done
   hdiutil detach /Volumes/CLIQZ
   SIGNED_DMG="${DMG%.dmg}-signed.dmg"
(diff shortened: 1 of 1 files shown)


sign_mac.sh
```diff
@@ -14,11 +14,16 @@ do
...
+      echo "Something went wrong during signing"
+      exit 1
+    fi
   done
   hdiutil detach /Volumes/CLIQZ
   SIGNED_DMG="${DMG%.dmg}-signed.dmg"
   appdmg CLIQZ-dmg.json $SIGNED_DMG
-  mv $SIGNED_DMG $DMG
+  echo "Replaced unsigned dmg with signed one"
+  cp -v $SIGNED_DMG $DMG
```
maxim-cliqz:
```
set -e
set -x
```

maxim-cliqz:
And please put issue ID into the commit message like this: "DB-XXX: Issue or fix summary"

sign_mac.sh
chrmod: :+1:

chrmod:
@maxim-cliqz right. was working without a ticket that's why haven't put it there

PR #191, opened 2016-02-22
DB-410: Duplicate notification line on addons internal page

maxim-cliqz:
Probably was introduced by the recent rewrite of git history.


luciancor:
retest this please

PR #192, opened 2016-02-23
Merge with FF 44.0.2

alver-cliqz:
I think it is better to merge with master after 0.9.3 release


chrmod:
@alver-cliqz have you check if german translations were updated?

alver-cliqz:
Yes. nothing changes between 44.0 and 44.0.2

PR #196, opened 2016-02-24
Build: Encapsulating shared part of magic scripts

chrmod:
For developers convinience we can just:
`source cliqz_env.sh`
to have all environment variables set.


maxim-cliqz:
Please specify ticked id in the commit message.

PR #204, opened 2016-03-01
DB-108: replace Mozilla Maintenance Service with Cliqz Service

alver-cliqz:
ServicesHelper.dll was replace with our build of this library, because need to change registry path in it.

All other text, related to this service, replaced to create a separate service, with similar functionality, as Mozilla's had.


alver-cliqz:
Sorry, make an error in naming in previous PR branch.

PR #206, opened 2016-03-01
DB-424 Force extension signature check on startup

chrmod:


cliqz.cfg
```diff
@@ -18,6 +21,25 @@ let extensionUrls = {
 
 let prefs = Cc["@mozilla.org/preferences-service;1"].getService(Ci.nsIPrefBranch);
 
+function verifyAddons() {
+  Cu.import("resource://gre/modules/addons/XPIProvider.jsm");
+  let versionChecker = Cc["@mozilla.org/xpcom/version-comparator;1"]
+                         .getService(Components.interfaces.nsIVersionComparator);
```
maxim-cliqz: Weird (and wasteful) indentation

cliqz.cfg
```diff
@@ -18,6 +21,25 @@ let extensionUrls = {
...
+  let distributionVersion = prefs.getCharPref("distribution.version");
+
+  let lastDistributionVersion;
+  try {
```
maxim-cliqz: Can it really throw?

cliqz.cfg
```diff
@@ -18,6 +21,25 @@ let extensionUrls = {
...
+    lastDistributionVersion = prefs.getCharPref("distribution.lastVersion");
+  } catch(e) {
+    lastDistributionVersion = 0;
+  }
+
+  if (versionChecker.compare(distributionVersion, lastDistributionVersion) > 0) {
+    XPIProvider.verifySignatures();
+    prefs.setCharPref("distribution.lastVersion", distributionVersion);
```
maxim-cliqz: There is "extensions.lastPlatformVersion" which serves the same purpose and is already maintained by `AddonManager`.
And actually I think it's best to do it where all other similar job is done – in `AddonManagerInternal.startup`

luciancor: temporary solution which should be removed/changed with a better one soon

cliqz.cfg
luciancor: the pref API is quite strange - if the pref is not there or not ready it simply throws an exception :/

PR #209, opened 2016-03-02
DB-450: update to latest extension 1.2.8

luciancor:


alver-cliqz:
Problem with signing on Mac (The timestamp service is not available.). Can ignore this and merge

PR #212, opened 2016-03-03
Typo fix in about:license

chrmod:
reported by Felix


alver-cliqz:
Windows PR builder restarted while upload to S3, but everything built normally. Can count as everything ok.

PR #225, opened 2016-03-10
Cleanup addons leftovers

maxim-cliqz:


maxim-cliqz:
retest this please

mozilla-release/toolkit/mozapps/extensions/internal/XPIProvider.jsm
```diff
@@ -255,6 +255,11 @@ const COMPATIBLE_BY_DEFAULT_TYPES = {
   dictionary: true
 };
 
+// A set of addon IDs which are allowed to run in safe mode.
+const gSafeModeWhitelistedAddons = new Set([
+  "cliqz@cliqz.com"
```
chrmod: we want to allow all `cliqz.com` addons. ATM there is only one `userassessment@cliqz.com`

maxim-cliqz: >we want to allow all cliqz.com
Why do you think so? I made an exception for one because it's an essential part of the browser.
But generally addons are disabled in safe mode for a reason.
Besides, `userassessment@cliqz.com` is an internal tool, not intended to be used outside. So I see no reason making exception for anything else.

chrmod: you're right

PR #233, opened 2016-03-16
DB-481: Research on addons blocking. DO NOT MERGE!

maxim-cliqz:
@chrmod @luciancor @alver-cliqz 
This is the code I finished working on before going on vacation.
I experimented with several ways to block addons.
The perfect one would be to only allow those that have our signature, but I didn't yet manage to get that working. It should probably be done somewhere in `XPIProvider` around `startup` and `loadManifest` functions.

### DIFF
--- a/mozilla-release/toolkit/mozapps/extensions/nsBlocklistService.js
+++ b/mozilla-release/toolkit/mozapps/extensions/nsBlocklistService.js
@@ -9,9 +9,11 @@
 const Cc = Components.classes;
 const Ci = Components.interfaces;
 const Cr = Components.results;
+const Cu = Components.utils;
 
 Components.utils.import("resource://gre/modules/XPCOMUtils.jsm");
 Components.utils.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/Console.jsm");
 
 try {
   // AddonManager.jsm doesn't allow itself to be imported in the child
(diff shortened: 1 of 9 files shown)


mozilla-release/toolkit/mozapps/extensions/nsBlocklistService.js
```diff
@@ -381,6 +383,14 @@ Blocklist.prototype = {
 
   /* See nsIBlocklistService */
   getAddonBlocklistState: function Blocklist_getAddonBlocklistState(addon, appVersion, toolkitVersion) {
+#if 0
+    // It's also possible to block all addons through blocklist.
+    console.log("BLOCKLIST CHECK", addon.id, addon.type);
+    const isCliqzExtension =
+        (addon.type === "extension") && (addon.id === "cliqz@cliqz.com");
```
luciancor: @maxim-cliqz this would not work because anybody could call their addon cliqz@cliqz.com

maxim-cliqz: @luciancor I know. Just a proof of concept :)

luciancor:
@maxim-cliqz what about simply returning true in https://github.com/maxim-cliqz/browser-f/blob/DB-481_1/mozilla-release/toolkit/mozapps/extensions/internal/XPIProvider.jsm#L238 ? :)

maxim-cliqz:
@luciancor agree, that's the simplest solution at the moment. But I'm afraid, it won't protect from malware that modifies profile data to simulate extension installs.

luciancor:
@maxim-cliqz what about https://github.com/cliqz-oss/browser-f/pull/241 and creating a different ticket for the malware case?

PR #242, opened 2016-04-07
DB-481: Better addons blocking

maxim-cliqz:
Allow  unsigned dictionaries as they should be harmless.
Verify addons signatures on every startup synchronously.
Report addons installation attempts to own telemetry server.


maxim-cliqz:
Wrong branch name

PR #249, opened 2016-04-12
DB-488: builds expect MOZILLA_API_KEY

chrmod:
Imlements: https://cliqztix.atlassian.net/browse/DB-488?jql=text%20~%20%22geolocation%22

Should not be merge until build servers will provide MOZ_MOZILLA_API_KEY as environment variable.

### DIFF
--- a/magic_build_and_package.sh
+++ b/magic_build_and_package.sh
@@ -7,6 +7,7 @@
 # Required ENVs:
 # WIN32_REDIST_DIR
 # CQZ_GOOGLE_API_KEY or MOZ_GOOGLE_API_KEY
+# MOZ_MOZILLA_API_KEY
 # CQZ_RELEASE_CHANNEL or MOZ_UPDATE_CHANNEL
 # CQZ_CERT_DB_PATH
 # MOZ_UI_LOCALE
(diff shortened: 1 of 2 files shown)


mozilla-release/browser/app/profile/firefox.js
```diff
@@ -1441,11 +1441,7 @@ pref("plain_text.wrap_long_lines", true);
 pref("dom.debug.propagate_gesture_events_through_content", false);
 
 // The request URL of the GeoLocation backend.
-#ifdef RELEASE_BUILD
-pref("geo.wifi.uri", "https://www.googleapis.com/geolocation/v1/geolocate?key=%GOOGLE_API_KEY%");
-#else
```
alver-cliqz: Did we want to use google geoservice as fallback solution?

magic_build_and_package.sh
```diff
@@ -27,6 +28,10 @@ if [ $CQZ_GOOGLE_API_KEY ]; then
   export MOZ_GOOGLE_API_KEY=$CQZ_GOOGLE_API_KEY  # --with-google-api-keyfile=...
 fi
 
+if [ -z $MOZ_MOZILLA_API_KEY ]; then
+  echo "warning: MOZ_MOZILLA_API_KEY environment variable is missing"
```
maxim-cliqz: If it's **required**, then the script should `exit 1` in case of its absence.

chrmod: I didn't do it, to allow people build without our API key.

PR #252, opened 2016-04-12
Merge with FF 45.0.1

alver-cliqz:


luciancor:
@alver-cliqz shouldn't we get 45.0.2 directly?

maxim-cliqz:
No conflicts at all?

alver-cliqz:
Yep, no conflicts.

And yes - better to merge with 45.0.2 directly, will do today

PR #254, opened 2016-04-13
Merge with FF 45.0.2

alver-cliqz:


maxim-cliqz:
Just a notice: those commits should have been marked as [minor], not [major], probably.

maxim-cliqz:
LGTM

alver-cliqz:
This is from my previous experience - mark all upstream releases as [major]. How do you think, do we need such kind of mark at all?

PR #261, opened 2016-04-19
DB-527: Disable Mozilla's tracking protection

maxim-cliqz:


mozilla-release/modules/libpref/init/all.js
```diff
@@ -1115,7 +1115,7 @@ pref("privacy.donottrackheader.enabled",    false);
 // Enforce tracking protection in all modes
 pref("privacy.trackingprotection.enabled",  false);
 // Enforce tracking protection in Private Browsing mode
-pref("privacy.trackingprotection.pbmode.enabled",  true);
+pref("privacy.trackingprotection.pbmode.enabled",  false);
```
chrmod: should we lock this one as well?

maxim-cliqz: it is locked – in cliqz.cfg

PR #267, opened 2016-04-25
DB-471: Skip default browser prompt during first run

maxim-cliqz:


mozilla-release/browser/components/nsBrowserGlue.js
```diff
@@ -1278,7 +1278,7 @@ BrowserGlue.prototype = {
       let shouldCheck = ShellService.shouldCheckDefaultBrowser;
 #endif
       let promptCount = 0;
-#ifndef RELEASE_BUILD
+#ifdef RELEASE_BUILD
```
alver-cliqz: Max, can you add a little comment about this change? Because promptCount stay in another ifdef block can be accidenly changed in future. Or let's create another protection from such kind of changes.

maxim-cliqz: Done.

alver-cliqz: thanks

PR #289, opened 2016-05-10
DB-624: make cliqz@cliqz.com a System Addon

chrmod:


chrmod:
tested it on windows and linux, worked just fine.

We still need to wait for balrog update, but browser should be ready

luciancor:
@chrmod is the mac builder down?

luciancor:
retest this please

alver-cliqz:
No, we still can face with problem of different xpi while building on Mac (because building doing twice).

PR #298, opened 2016-05-20
[WIP] DB-588: Linux build driven via Jenkinsfile and Docker

chrmod:
Also, we do submit build do APT repository


alver-cliqz:
Looks like you put files with HTTPS everywhere in this commit again

PR #301, opened 2016-05-24
Build: setting right repository

chrmod:


Jenkinsfile
```diff
@@ -7,10 +7,10 @@ node {
   } catch (MissingPropertyExceptionmpe) {
     CQZ_COMMIT = ''
   }
-  CQZ_COMMIT = CQZ_COMMIT ? CQZ_COMMIT : 'DB-588_2'
+  CQZ_COMMIT = CQZ_COMMIT ? CQZ_COMMIT : 'master'
```
alver-cliqz: May be better cancel build if CQZ_COMMIT doesn't set up?

PR #306, opened 2016-05-30
DB-652: Private tab indication

maxim-cliqz:
Paint urlbar background grey, when a private tab is active.

### DIFF
--- a/mozilla-release/browser/base/content/browser.js
+++ b/mozilla-release/browser/base/content/browser.js
@@ -235,6 +235,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "LoginManagerParent",
 #if CQZ_AUTO_PRIVATE_TAB
 XPCOMUtils.defineLazyModuleGetter(this, "AutoPrivateTab",
   "resource:///modules/AutoPrivateTab.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PrivateTabUI",
+  "chrome://browser/content/PrivateTabUI.jsm");
 #endif
 
 var gInitialPages = [
@@ -1010,6 +1012,10 @@ var gBrowserInit = {
     CombinedStopReload.init();
     gPrivateBrowsingUI.init();
     TabsInTitlebar.init();
+#if CQZ_AUTO_PRIVATE_TAB
+    this._privateTabUI = new PrivateTabUI(gBrowser, gNavToolbox);
+    this._privateTabUI.start();
+#endif
 
     if (window.matchMedia("(-moz-os-version: windows-win8)").matches &&
         window.matchMedia("(-moz-windows-default-theme)").matches) {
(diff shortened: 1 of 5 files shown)


mozilla-release/browser/base/content/browser.js
```diff
@@ -7396,6 +7399,44 @@ var gRemoteTabsUI = {
   }
 };
 
+// Small utility handler which propagates current tab "privateness" property as
+// available to CSS selectors through #navigator-toolbox[on_private_tab]
+const _gPrivateTabUI = {
```
chrmod: uh, I know Firefox tends to have big files, but can we at least do not add our stuff into them? Is there some way of modularization?

mozilla-release/browser/base/content/browser.js
```diff
@@ -7396,6 +7399,44 @@ var gRemoteTabsUI = {
...
+  init: function PrivateTabUI_init() {
+    this._updateNavToolbox();
+    for (let eventName in this._tabContainerEvents) {
```
chrmod:
```
Object.keys(this._tabContainerEvents).forEach( eventName => gBrowser.tabContainer.addEventListener(eventName, this) );
```

mozilla-release/browser/base/content/browser.js
```diff
@@ -7396,6 +7399,44 @@ var gRemoteTabsUI = {
...
+      gBrowser.tabContainer.addEventListener(eventName, this);
```
chrmod: does this work correctly? should it be
```
gBrowser.tabContainer.addEventListener(eventName, this._tabContainerEvents[eventName]);
```

mozilla-release/browser/base/content/browser.js
```diff
@@ -7396,6 +7399,44 @@ var gRemoteTabsUI = {
   }
 };
 
+// Small utility handler which propagates current tab "privateness" property as
+// available to CSS selectors through #navigator-toolbox[on_private_tab]
+const _gPrivateTabUI = {
+  init: function PrivateTabUI_init() {
```
chrmod: cannot we have a constructor? you could use **class** syntax

maxim-cliqz: In this particular case I'd find it inappropriate.
We need a corresponding "destructor" called at the right moment, and having `new X()` + `deinit` looks less obvious to me than a pair of `init` - `deinit`.

mozilla-release/browser/base/content/browser.js
maxim-cliqz: What's the benefit of doing it this way?

mozilla-release/browser/base/content/browser.js
maxim-cliqz: It works correctly. Object passed to `addEventListener ` is called through `handleEvent ` function.
Inside `handleEvent ` we pass control to appropriate handler calling it with context of `_gPrivateTabUI`, because `_tabContainerEvents` here is more a map container than an individual object.

mozilla-release/browser/base/content/browser.js
maxim-cliqz: There's some. 
I was too lazy :) https://en.wikipedia.org/wiki/Broken_windows_theory
Will attempt to fix this.

PR #309, opened 2016-06-02
DB-666: some improvement for new build system

alver-cliqz:
CQZ_BUILD_ID. It is hard to totally replace FF "buildid" because it spread all over the build scripts and expected exactly in format "%Y%m%d%H%M%S". We need to use same format in our build. Jenkins must generate environment variable CQZ_BUILD_ID (or can be set up manually on dev environment)

CQZ_BUILD_DE_LOCALIZATION. Optional, must be set up if we need in DE build (build with language repack)


chrmod:
:+1:

PR #314, opened 2016-06-06
Db 680 1

stefanie-cliqz:


l10n/de/browser/chrome/browser/aboutPrivateBrowsing.dtd
```diff
@@ -5,11 +5,11 @@
 <!ENTITY % brandDTD SYSTEM "chrome://branding/locale/brand.dtd">
```
maxim-cliqz: We don't use strings from this file anymore, since we replaced Mozilla's page in DB-662.
Can also safely leave these changes though.

l10n/de/mobile/android/base/android_strings.dtd
```diff
@@ -91,7 +91,7 @@
      We can't use android plural forms, sadly. See bug #753859. -->
```
maxim-cliqz: This file belongs to android version, as you see from its path and name.

l10n/de/suite/chrome/common/utilityOverlay.dtd
```diff
@@ -22,7 +22,7 @@
 <!ENTITY newNavigatorCmd.label "Browser-Fenster">
```
maxim-cliqz: "suite" directory is probably for the SeaMonkey, we don't need it now.

mozilla-release/toolkit/locales/en-US/chrome/mozapps/downloads/downloads.properties
```diff
@@ -32,14 +32,14 @@ offlineCancelDownloadsAlertTitle=Cancel All Downloads?
 offlineCancelDownloadsAlertMsg=If you go offline now, 1 download will be canceled. Are you sure you want to go offline?
```
maxim-cliqz: It's a bit suspicious that this file has no changes in German locale. Could you miss it?

stefanie-cliqz: I checked. It should be correct: 
offlineCancelDownloadsAlertMsg=Wenn du jetzt offline gehst, wird 1 Download abgebrochen. Bist du sicher, dass du offline gehen möchtest?

maxim-cliqz:
Fixed conflicts with latest Firefox and removed diffs in unused files. New PR is here: https://github.com/cliqz-oss/browser-f/pull/326

PR #330, opened 2016-06-14
DB-689: replace images for Forget mdoe on Windows

alver-cliqz:


alver-cliqz:
Still under development

PR #391, opened 2016-09-26
DB-789: add "reset to default" function for search engines list on about:preferences#search

mai-cliqz:


alver-cliqz:
retest this please

PR #407, opened 2016-10-04
EX-2647: Add startpage.com search to list of default alternate search engines in CLIQZ

mai-cliqz:


alver-cliqz:
This commit add startpage.com only into the browser. As i see in ticket description - we want to have it in separate extension too. May be this must be implement in different way (in extension itself, not in the browser)?

mai-cliqz:
@alver-cliqz We need to implement this startpage in both browser and extension in order to make "start page" as a default search engine. If we implement it only on the extension, when user Remove it, they cannot Restore it anymore.

PR #410, opened 2016-10-05
DB-765: AFT attempts cleanup only if domain was extracted from URL

maxim-cliqz:


mozilla-release/browser/components/AutoPrivateTab.js
```diff
@@ -335,7 +335,13 @@ AutoPrivateTabDatabase.prototype = {
     tab.private = true;
     tabBrowser.reload();
 
-    if (remember) {
+    if (domain) {
```
maxim-cliqz: Fur internal pages with URLs like "about:*" or "chrome:*" the domain is undefined, so we shouldn't even attemp to cleanup anything based on it, hence no user prompt is needed as well.

PR #429, opened 2016-10-27
DB-970: update extension to 1.10.0

luciancor:
@alver-cliqz can we please have a new version 1.8.1 of the browser containing this update?


alver-cliqz:
Sure. We are not using this links anymore with new build environment. I will create a new version

PR #432, opened 2016-10-31
DB-993: fix PGO build

alver-cliqz:
Do not show onboarding from CLIQZ extension on build process only


alver-cliqz:
Checked locally with release build - works as expected

PR #444, opened 2016-11-15
DB-985: Fix and complete implementation of Firefox Data import into Cliqz browser

mai-cliqz:
@luciancor

### DIFF
--- a/mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
+++ b/mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
@@ -18,6 +18,8 @@ const { classes: Cc, interfaces: Ci, utils: Cu } = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource:///modules/MigrationUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
+Cu.import("resource://gre/modules/NetUtil.jsm");
 
 XPCOMUtils.defineLazyModuleGetter(this, "PlacesBackups",
                                   "resource://gre/modules/PlacesBackups.jsm");
@@ -31,6 +33,12 @@ XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge",
                                   "resource://gre/modules/ProfileAge.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "AppConstants",
                                   "resource://gre/modules/AppConstants.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PlacesUtils",
+                                  "resource://gre/modules/PlacesUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "Sqlite",
+                                  "resource://gre/modules/Sqlite.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "FormHistory",
+                                  "resource://gre/modules/FormHistory.jsm");
 
 
 /* start CLIQZ helpers */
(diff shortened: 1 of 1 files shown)


mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
```diff
@@ -180,7 +241,262 @@ FirefoxProfileMigrator.prototype._getResourcesInternal = function(sourceProfileD
...
+            yield db.close();
+          }
+        }).then(() => aCallback(true),
+        e => { Cu.reportError(e); aCallback(false) });
+      }
+    };          
+  }.bind(this);
+
   let types = MigrationUtils.resourceTypes;
+  if (!this.startupOnlyMigrator && !MigrationUtils.isStartupMigration) {
```
maxim-cliqz: Expression `!this.startupOnlyMigrator` is always `true`.

mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
```diff
@@ -54,6 +62,59 @@ function getFFFolder() {
...
 }
 
+function insertWholeBookmarkFolder(db, aId, aGuid) {
+  let query = `SELECT b.id, h.url, COALESCE(b.title, h.title) AS title, 
+    b.type, k.keyword, b.dateAdded, b.lastModified, h.favicon_id 
+    FROM moz_bookmarks b 
+    LEFT JOIN moz_places h ON b.fk = h.id 
+    LEFT JOIN moz_keywords k ON k.id = b.keyword_id 
+    WHERE b.type IN (1,2) AND b.parent = ${aId}
+    ORDER BY b.position;`;
```
maxim-cliqz: Database schemas may change from version to version. Did you test it with older Firefox?

mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
mai-cliqz: Yes. You're right. It's always `true` because we define it here:
```
Object.defineProperty(FirefoxProfileMigrator.prototype, "startupOnlyMigrator", {
   get: () => false
});
```
So I left it like that in case we want to disable this migration option, we just need to change the `get: () => true`

mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
mai-cliqz: I didn't test with older version. But checked the changes of the schemes since firefox 3. They have removed the table: `moz_bookmarks_root`. Chromium uses it to load root bookmarks id, we don't. So this function could work with older database version also.

mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
```diff
@@ -54,6 +62,59 @@ function getFFFolder() {
...
+  let rows = yield db.execute(query);
+  let yieldCounter = 0;
+  for (let row of rows) {
+    let type = row.getResultByName("type");
+    let title = row.getResultByName("title");
+    let id = row.getResultByName("id");
+
+    switch (type) {
+      case PlacesUtils.bookmarks.TYPE_BOOKMARK: // Bookmark Url - Handle keyword and favicon
```
maxim-cliqz: Please watch line lengths and the rest of coding conventions.

mozilla-release/browser/components/migration/FirefoxProfileMigrator.js
mai-cliqz: `this.startupOnlyMigrator` and `MigrationUtils.isStartupMigration` are different conditions:
`this.startupOnlyMigrator` indicates if Firefox is a startup migrator only or not. Firefox was a startup migrator only before but now it becomes a normal migrator also.
`MigrationUtils.isStartupMigration` indicates if this migrating action is executed right after installing the browser, or when they are using the browser.

PR #450, opened 2016-11-18
DB-989: HttpsEverywhere Enable/Disable to be removed from DB url menu…

mai-cliqz:
… (i)


mai-cliqz:
@luciancor please review it

PR #469, opened 2016-11-28
DB-1070: remove aushelper extension from CLIQZ browser

alver-cliqz:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1311515 it still not good plus we can not use it properly


maxim-cliqz:
retest this please

PR #471, opened 2016-11-28
DB-1041: some fixes for Windows installer

alver-cliqz:
Need to recheck:
1. Update to higher version
2. Installation on Windows 8
3. Make as default on silent installation


maxim-cliqz:
retest this please

PR #481, opened 2016-12-06
DB-911: Quick fixing of cross-window private tabs transplantation

maxim-cliqz:


mozilla-release/docshell/base/nsDocShell.cpp
```diff
@@ -3409,6 +3409,7 @@ nsDocShell::SetDocLoaderParent(nsDocLoader* aParent)
       // DocShells are swapped during tab being dragged between windows, and
       // we don't want private tabs to lose their privacy flag.
       // See DB-911.
+      // Note: this path seems obsolete after merge with Fx-50.
```
maxim-cliqz: @mikeconley Hi Mike. Could you take a look here? We ran into a weird problem with forget tabs after merging with Firefox-50: `nsILoadContext.usePrivateBrowsing` gets lost after DocShell swapping. Previously it was handled by this condition, and now I had to patch it in the `nsFrameLoader`. But that's not all. For a reason I absolutely do not understand `xul:browser`'s `loadContext` and `docShell` now point to different objects, while they're supposed to be the same instance of `nsDocShell` (single-process mode). For that reason I had to also patch-up the `usePrivateBrowsing` in the `tabbrowser.xml`, as it gets lost somehow.
As you see, it all gets ugly. If you could clarify what's going on, and how we can improve it, that'd be of a great help to us.

PR #483, opened 2016-12-09
DB-799: Add signals for Auto Forget Mode to Browser Telemetry

sebastiandziadzio:
What we measure:

- number of times the Auto Forget Tab notification was displayed
- number of times user picked "Reload in Normal Mode"
- number of times user picked "Always load in Normal Mode"
- number of times user picked "Change Settings"

@maxim-cliqz, @luciancor: that's my first piece of code in JS, so I'd appreciate a review :)


luciancor:
@sebastiandziadzio Looks good!

PR #497, opened 2016-12-29
DB-1118: Inform users that LastPass installation was successful

mai-cliqz:


mozilla-release/toolkit/mozapps/extensions/content/extensions.xml
```diff
@@ -1814,6 +1814,22 @@
...
+          function openPopup() {
+            var lpBtn = doc.getElementById("toggle-button--supportlastpasscom-lastpass-button");
+            if(lpBtn) {
+              lpBtn.click();
+              if(doc.getElementsByTagName('iframe').length > nIframe || nTry === 0) {
+                return;
+              }
+            }
+            nTry--;
+            timer.initWithCallback(openPopup, 500, Ci.nsITimer.TYPE_ONE_SHOT);
```
luciancor: @mai-cliqz if the button is never there this code will never stop running. Can you please move the nTry ===0 part here and only reschedule a check if nTry !== 0 ?

PR #547, opened 2017-02-27
DB-1225: Remove hyperlink from crash report IDs on about:crashes

mai-cliqz:


mai-cliqz:
Applying `link.setAttribute("disabled", true)` doesn't work in this case because every anchor tag in the list has either href or event listener or both on it. To solve the problem, we need to disable the anchor tag and remove the event listener. One alternative solution is using a CSS property `pointer-events: none;` to disable all the pointer events on these links.

PR #548, opened 2017-03-02
Beta channel update

leonidcliqz:


luciancor:
deployed!

Jenkinsfile.release
```diff
@@ -192,9 +192,39 @@ stage('Build') {
     sendEmail(BUILD_STATUS_FAILED)
     throw e
   }
+  if (CQZ_RELEASE_CHANNEL == "beta") {
+      uploadBeta(CQZ_BUILD_ID, CQZ_VERSION)
+  }
```
alver-cliqz: Why this block outside of try-catch?

leonidcliqz: because i only want to run it if build succeed

PR #573, opened 2017-03-10
DB-1255: Mozilla's requests plus a test

maxim-cliqz:
I suggest to wait with merging till Mozilla finally accepts our patch.


maxim-cliqz:
Run like this
`mochitest -f chrome docshell/test/chrome/test_docRedirect.xul`

PR #580, opened 2017-03-14
DB-1262: fix error in parsing svg file while open "About" dialog

alver-cliqz:


mozilla-release/browser/base/content/aboutDialog.css
```diff
@@ -11,7 +11,8 @@
 }
 
 #rightBox {
-  background-image: url("chrome://branding/content/about-wordmark.svg");
+  /* CLIQZ. Don't use
+  background-image: url("chrome://branding/content/about-wordmark.svg");*/
```
maxim-cliqz: Why don't we use it?

alver-cliqz: Please check ticket DB-1262, there are an image with it. It is out of our design for About dialog

PR #598, opened 2017-03-31
SystemAddons: do not uninstall them every time an update comes

luciancor:
@alver-cliqz @past


luciancor:
closing this for now as we have a fix in the extension

PR #600, opened 2017-04-25
Fixes for Linux builds:

philipp-cliqz:
- Upgraded gcc version (using Ubuntu 17.04 instead of 14.04)
- Fixed: Rust should be available for the non-root user

@chrmod

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:17.04
 
 RUN apt-get update && apt-get install -y \
   alien \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -40,4 +36,8 @@ RUN sed -i.bkp -e \
       's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' \
       /etc/sudoers
 
+RUN wget -O bootstrap.py https://hg.mozilla.org/mozilla-central/raw-file/default/python/mozboot/bin/bootstrap.py && \
```
chrmod: you have to add a `USER` directive to change the user

alver-cliqz:
We don't need this for now, fix was done in #601

PR #608, opened 2017-04-27
DB-1103: Change URL of & add telemetry signal to "Report deceptive we…

mai-cliqz:
…bsite" function

### DIFF
--- a/mozilla-release/toolkit/components/url-classifier/SafeBrowsing.jsm
+++ b/mozilla-release/toolkit/components/url-classifier/SafeBrowsing.jsm
@@ -153,12 +153,15 @@ this.SafeBrowsing = {
     switch (kind) {
       case "Phish":
         pref = "browser.safebrowsing.reportPhishURL";
+        Services.telemetry.getHistogramById("REPORT_DECEPTIVE_SITE").add(0);
         break;
       case "PhishMistake":
         pref = "browser.safebrowsing.reportPhishMistakeURL";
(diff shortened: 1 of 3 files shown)


mozilla-release/toolkit/components/url-classifier/SafeBrowsing.jsm
```diff
@@ -153,12 +153,15 @@ this.SafeBrowsing = {
     switch (kind) {
       case "Phish":
         pref = "browser.safebrowsing.reportPhishURL";
+        Services.telemetry.getHistogramById("REPORT_DECEPTIVE_SITE").add(0);
```
luciancor: @mai-cliqz what does 0, 1, and 2 mean in this case?

mai-cliqz: @luciancor (0, 1, 2) mean ('Phish', 'PhishMistake',  'MalwareMistake'). We have three kind of signals. Since we can only accumulate the signals by numbers, we need to map these strings in to integer values.

mai-cliqz: cc @sebastiandziadzio https://github.com/cliqz-oss/browser-f/pull/483/files

sebastiandziadzio: Could you maybe include that in the histogram description for future reference?

PR #657, opened 2017-07-25
DB-1404: fix linkedBrowser "We're in trouble" message.

alver-cliqz:
Remove a code, same as in original FF 54.0

### DIFF
--- a/mozilla-release/browser/base/content/tabbrowser.xml
+++ b/mozilla-release/browser/base/content/tabbrowser.xml
@@ -2161,13 +2161,6 @@
               this.mPanelContainer.appendChild(notificationbox);
             }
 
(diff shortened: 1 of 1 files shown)


alver-cliqz:
I did yesterday. This code was a part of original FF (till version 54), https://github.com/cliqz-oss/browser-f/blob/248653f68cfb30f3621e54b73ffab5881541da0d/mozilla-release/browser/base/content/tabbrowser.xml#L2086. In commit https://github.com/cliqz-oss/browser-f/blob/6acdba99f217693528de5636e3a68602bf413d71/mozilla-release/browser/base/content/tabbrowser.xml#L1873 Max moved this code after injection.

PR #658, opened 2017-07-28
DB-1284: using adult-domains from extension

alver-cliqz:
If it doesn't exist - use from installation folder. With this solution we can update list with updating extension only, no needs to update whole browser.

### DIFF
--- a/mozilla-release/browser/components/AutoForgetTabsService.js
+++ b/mozilla-release/browser/components/AutoForgetTabsService.js
@@ -6,6 +6,8 @@ const {classes: Cc, interfaces: Ci, utils: Cu, results: Cr} = Components;
 
 const XPC_ID = "{998eec19-ade7-4b42-ad3f-5d19d98de71d}";
 const ADULT_DOMAINS_BF_FILE_NAME = "adult-domains.bin";
+const ADULT_DOMAINS_BF_RESOURCE_PATH =
+    "chrome://cliqz/content/freshtab/adult-domains.bin";
 const USR_BLACKLIST_FILE_NAME = "apt-extra-domains.json";
 const USR_WHITELIST_FILE_NAME = "apt-white-domains.json";
 const NOTIFICATION_TIMEOUT_MS = 60000;  // 1 minute.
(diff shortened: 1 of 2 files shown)


mozilla-release/browser/components/AutoForgetTabsService.js
```diff
@@ -254,17 +256,35 @@ AutoForgetTabsService.prototype = {
...
+      stream = channel.open();
     }
-    if (bfFile.exists() && bfFile.isFile()) {
-      let [filter, version] = BloomFilterUtils.loadFromFile(bfFile);
-      this._adultDomainsBF = filter;
-      this._adultDomainsVer = version;
+    catch(e) { }
+
+    if (!stream) {
+      var bfFile = FileUtils.getFile("XCurProcD", [ADULT_DOMAINS_BF_FILE_NAME]);
```
chrmod: please use const instead of var

mozilla-release/browser/components/AutoForgetTabsService.js
```diff
@@ -254,17 +256,35 @@ AutoForgetTabsService.prototype = {
...
+          .getService(Ci.nsIIOService);
+      var adultDomainsURI = ios.newURI(ADULT_DOMAINS_BF_RESOURCE_PATH);
+      var channel = ios.newChannelFromURIWithLoadInfo(adultDomainsURI, null);
+      stream = channel.open();
     }
-    if (bfFile.exists() && bfFile.isFile()) {
-      let [filter, version] = BloomFilterUtils.loadFromFile(bfFile);
-      this._adultDomainsBF = filter;
-      this._adultDomainsVer = version;
+    catch(e) { }
```
chrmod: some error reporting may be handy

mozilla-release/browser/components/AutoForgetTabsService.js
```diff
@@ -254,17 +256,35 @@ AutoForgetTabsService.prototype = {
   },
 
   _load: function AFTSvc_load() {
-    var bfFile = FileUtils.getFile("ProfD", [ADULT_DOMAINS_BF_FILE_NAME]);
-    if (!bfFile.exists() || !bfFile.isFile()) {
-      bfFile = FileUtils.getFile("XCurProcD", [ADULT_DOMAINS_BF_FILE_NAME]);
+    var stream;
+    try {
+      var ios = Cc["@mozilla.org/network/io-service;1"]
```
chrmod: Alternatively use `Serices.io`

mozilla-release/browser/components/AutoForgetTabsService.js
```diff
@@ -254,17 +256,35 @@ AutoForgetTabsService.prototype = {
   },
 
   _load: function AFTSvc_load() {
-    var bfFile = FileUtils.getFile("ProfD", [ADULT_DOMAINS_BF_FILE_NAME]);
-    if (!bfFile.exists() || !bfFile.isFile()) {
-      bfFile = FileUtils.getFile("XCurProcD", [ADULT_DOMAINS_BF_FILE_NAME]);
+    var stream;
```
chrmod: replace var with let please

chrmod:
generally great idea!

luciancor:
@chrmod we should keep this file in both places as the whole functionality is implemented in the browser and not being able to use it or test it without the extension doesn't seem right.

chrmod:
browser does not really work without the extension, so I don't see the point.  
In ideal world, the file should be handled by ResourceLoader. So maybe we should rely more on CLIQZ.jsm.

PR #659, opened 2017-07-31
DB-1412: use up to 4 process for e10s by default in Cliqz browser

alver-cliqz:


alver-cliqz:
@luciancor, is it what we want to turn on, right?

cliqz.cfg
```diff
@@ -42,6 +42,9 @@ lockPref("distribution.version", AppConstants.MOZ_APP_VERSION_DISPLAY);
 
 pref("browser.uitour.enabled", false);
 
+// Use up to 4 processes for webpages rendering (e10s feature)
```
luciancor: it is called e10s-multi - but yes, this should do the trick

alver-cliqz: yes, it do, i checked on my PC

chrmod:
we should start to move on from `cliqz.cfg`, to some more standard location.

@alver-cliqz how about `mozilla-release/browser/app/profile/firefox.js` ?

alver-cliqz:
@chrmod i don't like idea to mix with original FF files, because of merge. But we can discuss this, what we can improve here.

PR #674, opened 2017-08-25
EX-5396: new Cliqz Tab url

chrmod:


mozilla-release/browser/base/content/browser.js
```diff
@@ -297,7 +297,7 @@ var gInitialPages = [
   "about:privatebrowsing",
   "about:welcomeback",
   "about:sessionrestore",
-  "resource://cliqz/fresh-tab-frontend/index.html",
+  "resource://cliqz/freshtab/home.html",
```
alver-cliqz: What about old browsers?

alver-cliqz:
What about old browsers? Users will see something strange in URL bar till they updated to latest Cliqz browser version?

PR #677, opened 2017-08-25
DB-1433: Customize. Restore Defaults remove Cliqz icon from Toolbar

mai-cliqz:


mozilla-release/browser/components/customizableui/CustomizableUI.jsm
```diff
@@ -246,6 +246,7 @@ var CustomizableUIInternal = {
       "cliqz-vd-btn", // Cliqz video downloader 
       "control-center-browser-action", // Cliqz control center
       "cliqz-cc-btn", // Cliqz control center old
+      "mobilepairing_btn", // Cliqz connect/pairing
```
luciancor: this should not be in the nav bar

PR #680, opened 2017-08-29
DB-1458: fix reference error

alver-cliqz:


alver-cliqz:
It was allow to use in XML before without import. This is our part of code, Max did this.

PR #710, opened 2017-10-06
DB-1428: trigger System Addon extensions update check

alver-cliqz:
Together with usual browser update checking


luciancor:
@alver-cliqz we could also add some UI indicator but for now this should be sufficient

alver-cliqz:
@luciancor yes, we discussed this moment with Oleksandra already.

PR #715, opened 2017-10-09
Fixing mac signing

leonidcliqz:


Jenkinsfile
```diff
@@ -187,13 +212,12 @@ jobs["mac"] = {
...
             stage('OSX Bootstrap') {
-                sh '''#!/bin/bash -lc -x
-                rvm use ruby-2.4.2
-                pip install compare-locales
-                python mozilla-release/python/mozboot/bin/bootstrap.py --application-choice=browser --no-interactive
-                brew uninstall terminal-notifier
-                brew install wget --with-libressl
-                '''
+                withRVM('ruby-2.4.2') {
+                    sh '/bin/bash -lc "pip install compare-locales"'
```
chrmod: single `sh` block with `set -e` would probably work better

Jenkinsfile
```diff
@@ -187,13 +212,12 @@ jobs["mac"] = {
...
+                    sh '/bin/bash -lc "python mozilla-release/python/mozboot/bin/bootstrap.py --application-choice=browser --no-interactive"'
+                    sh '/bin/bash -lc "brew uninstall terminal-notifier"'
```
chrmod: wonder if we still need it, it was a problem 2 years ago

Jenkinsfile
```diff
@@ -12,6 +12,28 @@ CQZ_BUILD_ID = new Date().format('yyyyMMddHHmmss')
 def jobs = [:]
 def helpers
 
+def withRVM(version, cl) {
```
chrmod: how about putting it in our shared library ? https://github.com/cliqz-oss/jenkins-pipeline-library

Jenkinsfile
```diff
@@ -12,6 +12,28 @@ CQZ_BUILD_ID = new Date().format('yyyyMMddHHmmss')
...
+
+    withEnv(["PATH=${env.PATH}:$RVM_HOME", "RVM_HOME=$RVM_HOME"]) {
+        sh "set +x; source $RVM_HOME/scripts/rvm; rvm use --create --install --binary $version"
+    }
+
+    withEnv([
+        "PATH=$path",
+        "RUBY_VERSION=$version"
+        ]) {
+            cl()
```
chrmod: please fix the indent

Jenkinsfile
```diff
@@ -12,6 +12,28 @@ CQZ_BUILD_ID = new Date().format('yyyyMMddHHmmss')
...
 def helpers
 
+def withRVM(version, cl) {
+    RVM_HOME='$HOME/.rvm'
+    paths = [
+        "$RVM_HOME/bin",
+        "${env.PATH}"
+    ]
+    
+    def path = paths.join(':')
```
chrmod: in this case I'm not sure if this syntax is more readable, how about: `def path = "${RVM_HOME}/bind:${env.PATH}"`

PR #748, opened 2017-11-17
DB-1547: fix phishing sites reporting

alver-cliqz:


mozilla-release/toolkit/components/url-classifier/SafeBrowsing.jsm
```diff
@@ -191,19 +191,19 @@ this.SafeBrowsing = {
       case "Phish":
         pref = "browser.safebrowsing.reportPhishURL";
         Services.telemetry.getHistogramById("REPORT_DECEPTIVE_SITE").add(0);
-        reportPhishingURL(URI.asciiSpec, 'user-report-phish');
+        reportPhishingURL(info.uri, 'user-report-phish');
```
chrmod: are you sure this is correct? `asciiSpec` is a string, while URIs are typically objects..

alver-cliqz: I checked in debugger - "uri" is string in this "info" object.

PR #789, opened 2017-11-30
DB-1602: merge with FF 57.0.1

alver-cliqz:


mozilla-release/browser/components/nsBrowserGlue.js
```diff
@@ -1731,7 +1731,7 @@ BrowserGlue.prototype = {
 
   // eslint-disable-next-line complexity
   _migrateUI: function BG__migrateUI() {
-    const UI_VERSION = 58;
+    const UI_VERSION = 59;
```
chrmod: is this alright?

alver-cliqz: yes, this is how Mozilla processing migration between version. This is not related to browser version, this is internal variable which they change, when they need to do some fixes with settings.

PR #793, opened 2017-12-14
DB-1335, DB-1564: improve build scripts

alver-cliqz:
1. Remove outdated stuff
2. Add support for 64-bit build on Windows

From now it's possible just to start build_win.bat (Windows) or magic_build_and_package.sh (MacOS, Linux) to build a project with default settings (no needs to specify any variable)

### DIFF
--- a/cliqz_env.sh
+++ b/cliqz_env.sh
@@ -55,34 +55,43 @@ if [ "$CQZ_RELEASE_CHANNEL" = "release" ]; then
   export MOZ_UPDATE_CHANNEL=release
   # turn on PGO only for Release Windows build
   if [ $IS_WIN ]; then
-    if [ $CQZ_BUILD_ID ]; then
-      export MOZ_PGO=1 # release build optimization flag
-    fi
+    export MOZ_PGO=1 # release build optimization flag
   fi
 fi
 
 export MOZ_OBJDIR=../obj
 
-export MOZCONFIG=browser/config/cliqz-release.mozconfig
+if [ "$CQZ_BUILD_64BIT_WINDOWS" = "1" ]; then
+  export MOZCONFIG=browser/config/cliqz-release-64.mozconfig
+else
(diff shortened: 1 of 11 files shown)


cliqz_env.sh
```diff
@@ -55,34 +55,43 @@ if [ "$CQZ_RELEASE_CHANNEL" = "release" ]; then
...
-      export MOZ_PGO=1 # release build optimization flag
-    fi
+    export MOZ_PGO=1 # release build optimization flag
   fi
 fi
 
 export MOZ_OBJDIR=../obj
 
-export MOZCONFIG=browser/config/cliqz-release.mozconfig
+if [ "$CQZ_BUILD_64BIT_WINDOWS" = "1" ]; then
```
chrmod: shouldn't this be `==` ?

magic_upload_files.sh
```diff
@@ -10,7 +10,7 @@ cd $SRC_BASE
 cd $OBJ_DIR
 
 echo '***** Generate MAR for DE, if needed *****'
-if [ $CQZ_BUILD_DE_LOCALIZATION ]; then
+if [ "$CQZ_BUILD_DE_LOCALIZATION" = "1" ]; then
```
chrmod: `==`

cliqz_env.sh
luciancor: this should be the same in this case: 

[ STRING1 == STRING2 ] | True if the strings are equal.  "=" may be used instead of "==" for strict POSIX compliance.

http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

alver-cliqz: I did this in the same way, as I did before with checking for "release" channel (in same script). Can change everywhere to ==, let me check more carefully

alver-cliqz:
@chrmod @luciancor please check latest commit

luciancor:
Looks good!

PR #822, opened 2018-02-01
Removed nss installation from the build job

leonidcliqz:


chrmod:
And why does it work now? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

chrmod:
Instead of installing the nss from brew we could use one being build with the browser.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

alver-cliqz:
@chrmod just a try to resolve this temporary. We need a build. Later we will find a better solution. Not sure it will work - testing build in progress now

PR #835, opened 2018-02-07
[Do Not Merge] fix mach build tool to build "Android in Artifact Mode" using CLIQZ repository

saqib-cliqz:
Hi Team,

Currently we can't use mach to build "Android in Artifact Mode" using the CLIQZ repository. Because the tool makes certain assumptions e.g checkout from Mercurial, presence of Adjust SDK, which aren't true in our case.

The following changes needs to be done to make it work:

1. Hard-code urls to artifacts that needs to be downloaded. Normally, the tool uses Mercurial revisions to find out which version of the artifacts it's needs to download. As we don't use Mercurial, I would suggest to copy the required artifacts to a static url e.g AWS S3
2. The tool assumes the presence of Adjust SDK, which we are not using. Instead of killing the pipeline, I would recommend to silently ignore the error for the moment
3. Normally, mach should automatically detect that we are using GIT. However this detection is not working. In this case, I have hard-coded a default path which can be changed via env.

@alver-cliqz @spacifici
I would like to ask for your feedback regarding these changes. Can we solve these problems more elegantly?

### DIFF
--- a/mozilla-release/Makefile.in
+++ b/mozilla-release/Makefile.in
@@ -210,7 +210,8 @@ binaries::
 endif
 
 recurse_artifact:
-	$(topsrcdir)/mach --log-no-times artifact install
+	$(topsrcdir)/mach --log-no-times artifact install https://queue.taskcluster.net/v1/task/J-VPXVOPSqOzeAsiLDsTcQ/artifacts/public/build/target.apk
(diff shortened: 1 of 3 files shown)


alver-cliqz:
this changes must be first checked, before merge, with our current desktop browser build process, on all platforms

PR #837, opened 2018-02-08
Fix vagrant links

leonidcliqz:
https://github.com/hashicorp/vagrant/issues/6772

Same issue to resolve

### DIFF
--- a/Jenkinsfile.win
+++ b/Jenkinsfile.win
@@ -32,6 +32,8 @@ node(WIN_BUILD_NODE) {
 
     def helpers = load "artifacts/build-helpers.groovy"
 
+    sh "find .vagrant/machines/ -type d -name '*.vmx.lck' -print0 | xargs -0 rm -rf"
+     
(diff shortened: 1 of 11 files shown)


Jenkinsfile.win
```diff
@@ -32,6 +32,8 @@ node(WIN_BUILD_NODE) {
 
     def helpers = load "artifacts/build-helpers.groovy"
 
+    sh "find .vagrant/machines/ -type d -name '*.vmx.lck' -print0 | xargs -0 rm -rf"
```
chrmod: can you please add some comment, why do we need this?

alver-cliqz: on Win64 builder this file stay in system after "vagrant halt --force" command and bolck launching next build

chrmod:
Would it delete a lock file for currently running machine? What will happen in such case?


<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #855, opened 2018-03-12
DB-1604: Hide page action in forget tab/window

private-face:


mozilla-release/browser/components/newtab/aboutNewTabService.js
```diff
@@ -98,6 +98,11 @@ AboutNewTabService.prototype = {
   _overridden: false,
   willNotifyUser: false,
 
+  CLIQZ_NEW_TAB_URLS: [
+    "resource://cliqz/freshtab/home.html",
+    "about:cliqz"
+  ],
+
```
alver-cliqz: @chrmod is it right?

chrmod: looks good to me. we should definitely use resource:// url fro freshtab.

PR #899, opened 2018-06-04
DB-1754: return telemetry back

alver-cliqz:


konark-cliqz:
@luciancor @alver-cliqz  : Does it turn reporting for everyone? If a user has disabled it manually, it will respect it or overwrite it?

luciancor:
@konark-cliqz this only changes the default value -  will not touch any user pref

PR #912, opened 2018-06-19
Unify behavior of urlbar selection on click and double-click for Unix.

remusao:
This default behavior was initially changed to prevent a 16-years old
bug in SeaMonkey, which was already fixed in Firefox since 2007. See
following issues for more details:

* https://bugzilla.mozilla.org/show_bug.cgi?id=190615
* https://bugzilla.mozilla.org/show_bug.cgi?id=611162

This has been tested in the following way:

- Build from source on Ubuntu 16.04
- Tested artifact on Ubuntu and made sure that the behavior is consistent with MacOs


remusao:
Thank you very much @alver-cliqz!

PR #915, opened 2018-06-26
[WIP] Tor/ultraprivate mode

acatarineu:
This adds a system WebExtension ultraprivatemode@cliqz.com.xpi which can be built from https://github.com/cliqz/ultraprivatemode (make package).
Also, it adds Tor client binaries and resources to the Cliqz install directory, which are extracted from Tor Browser builds via the scripts in https://github.com/cliqz/ultraprivatemode/tree/master/scripts.
For now, these artifacts are uploaded to cdn.cliqz.com/browser-f/fun-demo/, but we can probably find a better place. Besides, it's still not clear we'll want to have a separate WebExtension vs integrating it into navigation-extension, but for now it's easier this way and should be working.
Also, applied a Firefox patch to solve crashes related to some proxy API. This is already there in 61+ and should be ported in the next Firefox ESR minor release.


acatarineu:
Closing this one and will split into smaller ones.

PR #923, opened 2018-07-04
DB-1765: use old system to upload build artifacts to S3 and prepare f…

alver-cliqz:
…iles for Balrog


alver-cliqz:
Using scripts from previous version. For next version we will need to make a changes in our CI to fit into new publishing format

PR #967, opened 2018-08-21
DB-1818: updated cliqz logo for about pages

aleksei-cliqz:
@chrmod @private-face @alver-cliqz @luciancor


alver-cliqz:
@aleksei-cliqz for which reason did you change filename? please, rename svg file and change only it in this PR

PR #1001, opened 2018-09-17
DB-1858: Move urlbar style changes from ui/window.es to browser theme

acatarineu:
This is what ui/window module is setting for each window. @luciancor @mai-cliqz Is this ok? Or should this be in a common css instead of one for each platform?


mai-cliqz:
We shouldn't change css direclty here, as it may cause some conflicts if someone update the browser theme. Normally what we do is updating the sass files (in this case you can modify the modules/theme/shared.scss) on the navigation-extension, make a PR to the extension repository to keep track the changes, then copy the content of the complied css (it generates 3 files for Linux/Osx and Windows respectively) from the build folder and override these corresponding files on this repository, to make sure we have no conflict.

acatarineu:
@mai-cliqz Updated css after https://github.com/cliqz/navigation-extension/pull/6699 was merged.

PR #1012, opened 2018-10-09
DB-1820: restoring engine shortcut / alias

aleksei-cliqz:
@alver-cliqz @luciancor @chrmod

### DIFF
--- a/mozilla-release/browser/components/preferences/in-content/search.js
+++ b/mozilla-release/browser/components/preferences/in-content/search.js
@@ -510,7 +510,7 @@ EngineStore.prototype = {
 
         // The search service removes the alias when an engine is hidden,
         // so clear any alias we may have cached before unhiding the engine.
-        e.alias = "";
+        // e.alias = "";
 
         this._engines.splice(i, 0, e);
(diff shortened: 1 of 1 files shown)


mozilla-release/browser/components/preferences/in-content/search.js
```diff
@@ -510,7 +510,7 @@ EngineStore.prototype = {
 
         // The search service removes the alias when an engine is hidden,
         // so clear any alias we may have cached before unhiding the engine.
-        e.alias = "";
+        // e.alias = "";
```
luciancor: @aleksei-cliqz why don't we want to clean the alias anymore?

aleksei-cliqz: Because if a user wants to restore a default search engines list after removal some items of it then we want him to see an alias (default search engine hashtag) immediately.
Cleaning it does not allow him/her to see a hashtag unless reload a page.

luciancor: 👍

PR #1013, opened 2018-10-10
Fix | download failed message on AMO - prevent mozaddonmanager

tsl143:
AMO uses `window.navigator.mozAddonManager` to handle addon installs/remove in Firefox. This leads to handling of failure messages by AMO not Firefox. Disabling `mozAddonManager` restores the native behaviour of browser on AMO, and failure messages are displayed as before.

### DIFF
--- a/cliqz.cfg
+++ b/cliqz.cfg
@@ -207,6 +207,7 @@ if (getenv('MOZ_CLIQZ_PRIVATE_MODE')) {
   pref("browser.pocket.api", "");
   pref("browser.pocket.site", "");
   pref("network.http.referer.hideOnionSource", true);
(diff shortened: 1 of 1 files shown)


cliqz.cfg
```diff
@@ -207,6 +207,7 @@ if (getenv('MOZ_CLIQZ_PRIVATE_MODE')) {
   pref("browser.pocket.api", "");
   pref("browser.pocket.site", "");
   pref("network.http.referer.hideOnionSource", true);
+  pref("privacy.resistFingerprinting.block_mozAddonManager", true); // Prevent mozaddonmanager on AMO | DB-1879
```
konarkmodi: `DB-1879` looks like an internal ticket. It is not very helpful here, since the ticket system is internal only.

PR #1047, opened 2018-11-23
DB-1760: restore in multiple windows fixed

aleksei-cliqz:
@alver-cliqz @tsl143 @luciancor

### DIFF
--- a/mozilla-release/browser/components/sessionstore/SessionStore.jsm
+++ b/mozilla-release/browser/components/sessionstore/SessionStore.jsm
@@ -3741,7 +3741,30 @@ var SessionStoreInternal = {
       return;
     }
 
-    let firstWindowData = root.windows.splice(0, 1);
+    // CLIQZ-SPECIAL:
+    // aWindow contains data which does not correspond to any session state yet.
+    // Also it contains a "page" information that is to be displayed in a loading tab,
+    // user could click on a link in some of his/her applications.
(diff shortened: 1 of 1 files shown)


mozilla-release/browser/components/sessionstore/SessionStore.jsm
```diff
@@ -3741,7 +3741,30 @@ var SessionStoreInternal = {
...
+    // CLIQZ-SPECIAL:
+    // aWindow contains data which does not correspond to any session state yet.
+    // Also it contains a "page" information that is to be displayed in a loading tab,
+    // user could click on a link in some of his/her applications.
+    // root.windows is an Array of windows data retrieved from a user's session.
+    // Here we need to find a window with zIndex equaled to 1.
+    // That means this window will be on top of the others after restore process is done.
+    // Putting a new aWindow data into that window with zIndex = 1 will let us display
+    // a new tabbrowser (a new tab) always on top (explicitly visible) for a user.
+    let firstWindowData = null;
```
tsl143: can we still initialize the `firstWindowData` with `root.windows.splice(0, 1)`, so as to save another if condition

aleksei-cliqz: I do not think so.
1) If we do then splice will take affect and explicitly remove 1 item from array before we start loop through it.
2) We do not even know whether this 0th item is the one with zIndex equaled to 1 BEFORE starting a loop.

PR #1050, opened 2018-11-26
DB-1897: Move recommended addons from preferences to about:addon

tsl143:
This PR removes all the addons from preferences page (password manager and ghostery)
We create a new section on `about:addons` - `Recomended Addons` and put all ghostery and password managers there.


tsl143:
@aleksei-cliqz this block of code was written by cliqz and not Mozilla, so this will never cause conflict :)

PR #1062, opened 2018-12-10
DB-1942: showConsoleLogs for beta channel enabled

aleksei-cliqz:
@alver-cliqz @luciancor @tsl143


aleksei-cliqz:
Should be implemented in Navigation Extension project.

PR #1063, opened 2018-12-11
DB-1941: Design about:addons page

tsl143:


mozilla-release/toolkit/themes/shared/extensions/extensions.inc.css
```diff
@@ -551,11 +551,16 @@ button.warning {
   background-color: var(--in-content-box-background);
 }
 
-.list:focus > .addon[selected] {
+/* CLIQZ-SPECIAL  mods for suppressing blue shadow in recommended addons*/
```
tsl143: @alver-cliqz , I am not sure if it is fine to change in this file, is it fine?

PR #1082, opened 2018-12-19
DB-1957: view source page broken

tsl143:


mozilla-release/toolkit/components/viewsource/content/viewSource-content.js
```diff
@@ -605,7 +605,7 @@ var ViewSourceContent = {
                               null, referrerPolicy, // referrer
                               null, null, // postData, headers
                               Services.io.newURI(baseURI), null, false);
-                              Services.scriptSecurityManager.getSystemPrincipal());
+                              Services.scriptSecurityManager.getSystemPrincipal();
```
alver-cliqz: this is not a proper fix

tsl143: @alver-cliqz there was an extra closing parenthesis left which caused JS execution failure, removing it gave me proper running build with view-source page working.
Am I missing something?

PR #1085, opened 2018-12-19
DB-1958: Hide cliqz extension pages from permissions popup in preferences

tsl143:


mozilla-release/browser/components/preferences/sitePermissions.js
```diff
@@ -231,6 +235,9 @@ var gSitePermissionsManager = {
       return;
     let l10nId = this._getCapabilityString(perm.capability);
     let p = new Permission(perm.principal, perm.type, perm.capability, l10nId);
+    // CLIQZ: Skip permissions for Cliqz extension pages
+    if (p.origin.includes(EXTENSION_ID))
```
luciancor: @tsl143 couldn't we check for the type of extension and hide if it's a system addon?

tsl143: We dont get addon/id/type in `perm` object, this maps only URL on which permission is available.
So we need to get URL information, more over, IMO we should move away from allowing special privileges for system addon but spcific cliqz addon, as we install directly https-everywhere from AMO now.

luciancor: cool, makes sense. Thanks!

PR #1109, opened 2019-01-11
DB-1979: removing about:addons unnecessary scroll and 

tsl143:


tsl143:
@alver-cliqz  / @aleksei-cliqz This is ready for complete review now.

PR #1136, opened 2019-02-04
DB-1939: set browser.privatebrowsing.apt pref false for every new profile

tsl143:
DB-1939: set `browser.privatebrowsing.apt` pref false for every new profile


mozilla-release/browser/components/nsBrowserGlue.js
```diff
@@ -2301,6 +2301,9 @@ BrowserGlue.prototype = {
       // This is a new profile, nothing to migrate.
       Services.prefs.setIntPref("browser.migration.version", UI_VERSION);
 
+      // CLIQZ-SPECIAL: Default disabled AFT for new users.
+      Services.prefs.setIntPref("browser.privatebrowsing.apt", false);
```
aleksei-cliqz: @tsl143 Are you sure it works? Shouldn't it be Services.prefs.setBoolPref(...) instead?

tsl143: Already updated :P

aleksei-cliqz: @tsl143 great, thanks!

PR #1157, opened 2019-02-08
DB-2038: remove more activity-stream mentions

alver-cliqz:


mozilla-release/toolkit/modules/NewTabUtils.jsm
```diff
@@ -11,16 +11,18 @@ ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.importGlobalProperties(["btoa", "URL"]);
 
 // Android tests don't import these properly, so guard against that
+let didSuccessfulImport = false;
+#if 0
 let shortURL = {};
 let searchShortcuts = {};
-let didSuccessfulImport = false;
```
tsl143: lets not delete it to prevent conflict, it will not execute as its under if 0

alver-cliqz: I moved it higher

PR #1162, opened 2019-02-13
DB-1991: blank Cliqz tab while first start up | fixed

aleksei-cliqz:


mozilla-release/toolkit/mozapps/extensions/AddonManager.jsm
```diff
@@ -3874,6 +3901,7 @@ var AMTelemetry = {
 };
 
 this.AddonManager.init();
+this.AddonManager.isReadyAsync();
```
tsl143: We can get rid of this :)

aleksei-cliqz: @tsl143 You are right! Fixed.

PR #1164, opened 2019-02-13
Maillist update

alver-cliqz:


Jenkinsfile.release
```diff
@@ -11,7 +11,7 @@ CQZ_VERSION = ""
 BUILD_STATUS_FAILED = 'Failed'
 BUILD_STATUS_SUCCESSFUL = 'Successful'
 
-EMAIL_LIST = 'alexander@cliqz.com,leonid@cliqz.com,lucian@cliqz.com'
```
konark-cliqz: @luciancor  @alver-cliqz : May be a better idea to use a mailing list here instead of individual email IDs?

Jenkinsfile.release
```diff
@@ -11,7 +11,7 @@ CQZ_VERSION = ""
...
+EMAIL_LIST = 'alexander@cliqz.com,lucian@cliqz.com'
 EMAIL_LIST_SUCCESSFUL = 'kateryna@cliqz.com,prashant@cliqz.com,tamara@cliqz.com,marta@cliqz.com,oleksandra@cliqz.com'
```
konark-cliqz: Same as above comment here.

Jenkinsfile.release
luciancor: good point, @alver-cliqz Marty can help with that ;)

Jenkinsfile.release
alver-cliqz: yes, Marty already suggest this, will do soon

PR #1169, opened 2019-02-15
Add Dat extension as system addon

sammacbeth:
Adds the prototype dat extension as a system addon at build time. We will test this extension on the beta channel.

I've guessed from the other system addons how to do this, please let me know if other steps are required.

cc @chrmod


alver-cliqz:
What is it? I don't like an idea with putting a lot of extensions inside browser distribution. Each extension consume memory. Also - you need to add checking for beta/release channel, with this configuration this extension will be added to all builds.

sammacbeth:
We are testing [an extension](https://github.com/cliqz-oss/dat-webext) to add support for the `dat://` [p2p protocol](https://www.datprotocol.com/) in the browser (see also [beaker browser](https://beakerbrowser.com/)). This extension uses experimental APIs for the networking stack of the protocol, as well as for registering a custom protocol, therefore needs to be bundled as a system addon. If you are aware of another why to ship this without this requirement, please let me know.

Which flags indicate the release channel so I can add a check?

alver-cliqz:
Env. variable CQZ_RELEASE_CHANNEL must be "beta". Ok for beta, but later let's discuss before using in release

PR #1170, opened 2019-02-18
DB-2045: customization mode blank tab | fixed

aleksei-cliqz:


alver-cliqz:
This code exist since FF 55.0. Let's check more carefully, what's happened with it, so we need to comment it out

PR #1175, opened 2019-02-19
 DB-2048: add premissions to about:addons page / listing and details 

tsl143:


mozilla-release/toolkit/mozapps/extensions/content/extensions.js
```diff
@@ -2765,6 +2770,65 @@ var gDetailView = {
     document.getElementById("detail-icon").src = icon ? icon : "";
     document.getElementById("detail-creator").setCreator(aAddon.creator, aAddon.homepageURL);
 
+    //CLIQZ-SPECIAL: This adds permission area to addon-details
```
aleksei-cliqz: @tsl143 Would you like to put your peace of code into a separate function?
You could 've probably called it renderPermList or similar...

tsl143: I just feel like going with the flow of the function, its having all the view updates/render

PR #1187, opened 2019-02-26
DB-2065: Prevent installation of lang packs and themes

tsl143:


remusao:
@tsl143 Does this still allow to install dictionaries to get spell correction for different languages?

alver-cliqz:
@remusao no, this is only about langpack and themes

tsl143:
> @tsl143 Does this still allow to install dictionaries to get spell correction for different languages?

Yes user can install dictionaries.

@remusao

PR #1190, opened 2019-02-27
DB-2015: enabling dropdown for changing browser language in preferences

tsl143:


alver-cliqz:
This is not enough. Let's also put everything necessary into this commit also (languages, dictionaries) in build process. Because just with this settings some users can download language pack (selecting DE from EN version, for example) and browser will not work for them

PR #1191, opened 2019-02-28
DB-2057: blank page after profile refresh | fixed

aleksei-cliqz:


mozilla-release/browser/components/sessionstore/SessionHistory.jsm
```diff
@@ -1,466 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
alver-cliqz: Why you delete this file?

PR #1195, opened 2019-03-07
DB-2069: video downloader icon fixed

aleksei-cliqz:


aleksei-cliqz:
We have discussed with @luciancor to change that icon in Navigation Extension project. Not only change it but to adjust for both diff themes.
So I close this PR

PR #1233, opened 2019-04-03
DB-2111: browser bar turns into forget mode for AFM | fixed

aleksei-cliqz:


mozilla-release/browser/components/preferences/in-content/preferences.js
```diff
@@ -169,15 +169,18 @@ function gotoPref(aCategory) {
...
-  if (category === "paneConnect" && gConnectPane && gConnectPane.loadFrame) {
-    gConnectPane.loadFrame();
-  }
-
   // Updating the hash (below) or changing the selected category
   // will re-enter gotoPref.
   if (gLastHash == category && !subcategory)
     return;
 
+  // CLIQZ-SPECIAL:
```
tsl143: I am not sure if this should this be part of this commit

aleksei-cliqz: @tsl143 why not to put it here quickly?
Some benefits of that:
1) This can fix the problem.
2) This can fix it in the next upcoming release.

aleksei-cliqz: And
3. A fix itself is pretty simple to rely on ;)

tsl143: Because we follow ticket - commit mapping

PR #1243, opened 2019-04-11
DB-2126: adding whitelist for block auto play

tsl143:


mozilla-release/browser/components/nsBrowserGlue.js
```diff
@@ -2226,6 +2228,7 @@ BrowserGlue.prototype = {
     const BROWSER_DOCURL = AppConstants.BROWSER_CHROME_URL;
 
     let currentUIVersion;
+    PermissionsUtils.importFromPrefs("blockautoplay.", "autoplay-media");
```
alver-cliqz: will this code works on every migrate even if we did it before?

PR #1304, opened 2019-07-18
DB-2208: added a private attribute for tabs

aleksei-cliqz:
In this PR I added a "private" attribute for a tab xul element.
Whenever the extension requests a status of a tab first we check whether the tab has contents loaded into that (linkedBrowser.loadContext is not null). If it does not then we rely on the "private" attribute value. Otherwise we define privacy on linkedBrowser.loadContext.

### DIFF
--- a/mozilla-release/browser/base/content/browser.js
+++ b/mozilla-release/browser/base/content/browser.js
@@ -2646,6 +2646,7 @@ function BrowserOpenTab(event) {
       openTrustedLinkIn(BROWSER_NEW_TAB_URL, where, {
         relatedToCurrent,
         resolveOnNewTabCreated: resolve,
+        private: false,
       });
     }),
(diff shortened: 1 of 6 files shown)


mozilla-release/browser/base/content/browser.js
```diff
@@ -2646,6 +2646,7 @@ function BrowserOpenTab(event) {
       openTrustedLinkIn(BROWSER_NEW_TAB_URL, where, {
         relatedToCurrent,
         resolveOnNewTabCreated: resolve,
+        private: false,
```
tsl143: Please handle this for `BROWSER_NEW_TAB_URL` in forget window.

aleksei-cliqz: @tsl143 Handled.

PR #1312, opened 2019-07-23
DB-2225: remove unnecessary comment

alver-cliqz:


alver-cliqz:
I found only this one. All other in source code (which is fine, we follow Chromium style there for now)

PR #1314, opened 2019-07-24
DB-2208: temporary remove console message

alver-cliqz:


alver-cliqz:
@aleksei-cliqz we need to check solution with private flag more accurate for next update

PR #1323, opened 2019-07-29
DB-2234: added a check against nativeTab.linkedBrowser

aleksei-cliqz:


aleksei-cliqz:
I have closed the PR because of creating another one with improving the check against isBrowserPrivate.

PR #1351, opened 2019-09-17
DB-2272: url bar blurring of text | fixed

aleksei-cliqz:


aleksei-cliqz:
> But we need to keep check of this hack, if this module is moved to `.jsm` in future.

I agree. When we use Quantum bar eventually this hack will go away by itself.

PR #1365, opened 2019-10-23
DB-2307: remove open in new forget tab

tsl143:


l10n/de/browser/browser/menubar.ftl
```diff
@@ -18,7 +18,7 @@ menu-file-new-window =
     .label = ­Neues Fenster
     .accesskey = N
 menu-file-new-private-window =
-    .label = Neues privates Fenster
+    .label = Neues Vergessen Fenster
```
aleksei-cliqz: @tsl143 I am not sure this is correct. Could you please check again with a German native speaker?

alver-cliqz: Checked with Elisa, it's fine

tsl143: Thank you :)

PR #1366, opened 2019-10-24
DB-2309: Restore private mode address bar color

tsl143:


mozilla-release/browser/base/content/browser.css
```diff
@@ -1390,3 +1390,14 @@ toolbar[keyNav=true]:not([collapsed=true]):not([customizing=true]) toolbartabsto
 #wrapper-https-everywhere_cliqz_com-browser-action {
   display: none;
 }
+
+/* Cliqz - background color of urlbar in forget mode */
+[privatebrowsingmode] #nav-bar #urlbar-container #urlbar:-moz-lwtheme-darktext {
+  background-color: #BBC6D4 !important;
```
aleksei-cliqz: @tsl143 could you please avoid using important here? Adding important makes it harder to support the code afterwards.
Also I think it could be written like this
```
[privatebrowsingmode] #urlbar:-moz-lwtheme-darktext
```
#urlbar is unique since it is ID. I would suggest to simplify CSS selectors to make our life easier.
What do you think?

PR #1370, opened 2019-10-28
Unify behavior of urlbar selection on click and double-click for Unix.

remusao:
This default behavior was initially changed to prevent a 16-years old
bug in SeaMonkey, which was already fixed in Firefox since 2007. See
following issues for more details:

* https://bugzilla.mozilla.org/show_bug.cgi?id=190615
* https://bugzilla.mozilla.org/show_bug.cgi?id=611162

@alver-cliqz @aleksei-cliqz this is what I was mentioning last week. Note that this change was already done in this #912 but it seems that a recent merge might have reverted it. So this PR is exactly the same as before and simply brings back the behavior (nothing changed).

### DIFF
--- a/mozilla-release/browser/app/profile/firefox.js
+++ b/mozilla-release/browser/app/profile/firefox.js
@@ -277,6 +277,15 @@ pref("browser.warnOnQuit", true);
 pref("browser.fullscreen.autohide", true);
 pref("browser.overlink-delay", 80);
 
+// CLIQZ-SPECIAL: unify behavior on all platforms. Comment special case for
+// Linux systems and replace with two prefs below. Reason: This default behavior
+// was initially changed to prevent a 16-years old bug in SeaMonkey, which was
+// already fixed in Firefox since 2007. See following issues for more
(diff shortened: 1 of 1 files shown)


mozilla-release/browser/app/profile/firefox.js
```diff
@@ -277,16 +277,8 @@ pref("browser.warnOnQuit", true);
 pref("browser.fullscreen.autohide", true);
 pref("browser.overlink-delay", 80);
 
-#ifdef UNIX_BUT_NOT_MAC
```
tsl143: Lets not delete the lines, to avoid future merge conflicts.
We can wrap this in `#if 0` and below that define the prefs
Adding a comment will also help to avoid this in future merges.

remusao: I force-pushed to fix the commit with your suggestion. I did not yet try to compile locally (will do so now). Is it correct to assume that the preprocessor to evaluate these `#if` blocks support nested conditions?

PR #1371, opened 2019-10-28
DB-2291: removing unused checks from search autocomplete

tsl143:


alver-cliqz:
@private-face please take a look when you will have time (fyi)

PR #1392, opened 2019-11-25
DB-2338: improving system addon update

tsl143:
THis patch makes sure that addon packed with browser atleast gets checked against current installed version of system addon.

### DIFF
--- a/mozilla-release/toolkit/mozapps/extensions/internal/XPIInstall.jsm
+++ b/mozilla-release/toolkit/mozapps/extensions/internal/XPIInstall.jsm
@@ -3971,6 +3971,7 @@ var XPIInstall = {
       }
     } catch(e) {
       // :(
+        logger.warn('System addon update failed', e)
     }
   },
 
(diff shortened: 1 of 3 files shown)


mozilla-release/toolkit/mozapps/extensions/internal/XPIInstall.jsm
```diff
@@ -3943,6 +3943,37 @@ var XPIInstall = {
...
+      }
+
+      if (latestCliqz.version > cliqzFromSystem.version) {
+        let systemAddonLocation = XPIStates.getLocation(KEY_APP_SYSTEM_ADDONS);
+        await systemAddonLocation.installer.installAddonSet(
+          [latestCliqz]
+        );
+      }
+    } catch(e) {
+      // :(
```
aleksei-cliqz: @tsl143 could you please log the situation in case we fall down to the catch clause?
I think it might be easier for us to debug it.

PR #1401, opened 2019-12-10
Update default cookie policy to reject from unvisited

sammacbeth:
As discussed with @tsl143 , the reject tracker policy is currently not working properly as the tracking protection code has been removed. This change switches to a more protective default until we can implement a reject tracker policy using our own tracking list.

### DIFF
--- a/mozilla-release/browser/app/profile/firefox.js
+++ b/mozilla-release/browser/app/profile/firefox.js
@@ -1618,7 +1618,7 @@ pref("browser.ping-centre.production.endpoint", "");
 pref("media.gmp-provider.enabled", true);
 
 // Enable blocking access to storage from tracking resources by default.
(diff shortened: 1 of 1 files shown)


tsl143:
We still handle this pref from here

image

aleksei-cliqz:
> We still handle this pref from here
> 
> image

Since "reject tracker policy" does not work properly now may be it makes sense to remove that option temporarily from that "Type blocked" menu until it works again?
And then apply changes what @sam has made (BEHAVIOR_REJECT_UNVISITED by default is selected), what do you think @tsl143 ?

tsl143:
We can get rid of the option, but we need to make sure the existing users having value `4` fallback to `3`

aleksei-cliqz:
> We can get rid of the option, but we need to make sure the existing users having value `4` fallback to `3`

Then we can approve this PR and create a ticket about falling back to BEHAVIOR_LIMIT_FOREIGN in case of "reject tracker policy" is selected for old users.

PR #1411, opened 2019-12-30
DB-2354: fixed about cliqz dialog period char for DE

aleksei-cliqz:


tsl143:
Can you do the same for english version too, to keep this consistent

aleksei-cliqz:
> Can you do the same for english version too, to keep this consistent

Yes, sure. Done.

PR #1413, opened 2020-01-02
Set default cross-origin referer trimming to origin only

sammacbeth:
Changes referer policy for cross-origin requests (see [this page](https://wiki.mozilla.org/Security/Referrer) for description of pref value meanings).

This brings us into line with [Safari's latest changes](https://webkit.org/blog/9661/preventing-tracking-prevention-tracking/).


sammacbeth:
> Looks good, ideally no one depends on querystrings from referrer (but can there be any case of breakage?)

Maybe, but as Safari already has this as the default most sites should fix this. I expect Firefox to follow suit on this shortly too.

PR #1418, opened 2020-01-09
Add Dat settings

sammacbeth:
Adds a new preferences settings called 'Labs' (or 'Labor' in German) for us to put settings for incubating test features. The first of these is the Dat Protocol that this PR integrated via an iframe to the extension settings page (as with connect).

Screenshot 2020-01-09 at 12 05 48


sammacbeth:
Now rebasing against master as the `xul` files seem to now be gone in 72.

tsl143:
LGTM, we need to get DE translations before merge though. We also need newlines in few files

sammacbeth:
> LGTM, we need to get DE translations before merge though. We also need newlines in few files

I lost German translations in the rebase. Just pushed them again.

tsl143:
Also please add the new lines to files
l10n/de/browser/browser/preferences/preferences.ftl
mozilla-release/browser/themes/shared/incontentprefs/experiments.svg
mozilla-release/browser/locales/en-US/browser/preferences/preferences.ftl

mozilla-release/browser/components/preferences/in-content/experiments.js
```diff
@@ -0,0 +1,19 @@
+var { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
+
+var gExperimentsPane = {
+  _pane: null,
+
+  init: function ()
+  {
+    this._pane = document.getElementById("paneExperiments");
```
aleksei-cliqz: @sammacbeth Good job! The only thing is that I am not sure we need this._pane here. Since the element with id="paneExperiments" does not exist (at least I couldn't find it). Probably you have followed the same pattern we have for **privacy.js** / **connect.js** files. I have checked privacy section and this._pane is null there as well (we need to remove this._pane there in the future also).

tsl143: We need to do that for `connect.js` too - `_pane` looks unnecessary.

mozilla-release/docshell/base/nsDocShell.cpp
```diff
@@ -9180,6 +9180,11 @@ nsresult nsDocShell::InternalLoad(nsDocShellLoadState* aLoadState,
       if (uri->SchemeIs("moz-extension") && uri->GetSpecOrDefault().Equals(connectURLString)) {
         break;
       }
+      nsCString datURLString;
```
remusao: Should these lines be wrapped with a comment to indicate this is custom for Cliqz?

mozilla-release/browser/components/preferences/in-content/preferences.js
```diff
@@ -109,6 +109,15 @@ function init_all() {
   } catch (e) {
     Cu.reportError(e);
   }
+  if (typeof gExperimentsPane !== 'undefined') {
```
remusao: Should this be wrapped into a comment indicating that this is custom Cliqz code?

mozilla-release/browser/components/preferences/in-content/experiments.js
```diff
@@ -0,0 +1,19 @@
+var { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
```
remusao: `var` is so 2016!

mozilla-release/browser/components/preferences/in-content/experiments.js
tsl143: Done in https://github.com/cliqz-oss/browser-f/pull/1422

mozilla-release/docshell/base/nsDocShell.cpp
sammacbeth: This follows the Cliqz special code for connect that has a preceding comment

mozilla-release/browser/components/preferences/in-content/preferences.js
sammacbeth: Follows connect code.

sammacbeth:
I updated this as per suggestions from @konark-cliqz to have a hard enable and disable for the dat extension.

PR #1420, opened 2020-01-09
DB-2372: fixing system addon update request and enable flag

tsl143:


mozilla-release/toolkit/mozapps/extensions/internal/XPIInstall.jsm
```diff
@@ -4012,6 +4012,11 @@ var XPIInstall = {
   },
 
   async updateSystemAddons() {
+    const PREF_SYS_ADDON_UPDATE_ENABLED = "extensions.systemAddon.update.enabled";
+    if (!Services.prefs.getBoolPref(PREF_SYS_ADDON_UPDATE_ENABLED, true)) {
+      return false;
```
remusao: This method does not seem to expect any return value.
```suggestion
      return;
```

mozilla-release/toolkit/mozapps/update/UpdateService.jsm
```diff
@@ -3941,6 +3941,10 @@ Checker.prototype = {
       return;
     }
 
+    // CLIQZ-SPECIAL Start checking updates for System Addons also.
+    let { XPIProvider } = Components.utils.import("resource://gre/modules/addons/XPIProvider.jsm", {});
```
remusao: Nit.
```suggestion
    const { XPIProvider } = Components.utils.import("resource://gre/modules/addons/XPIProvider.jsm", {});
```

tsl143: Firefox uses `const` as DEFINE only.

remusao: #OldSchool :smile:

PR #1425, opened 2020-01-13
DB-2378: fixing moz-ext URL in url bar

tsl143:
With this check
image

Without this check
image

### DIFF
--- a/mozilla-release/browser/components/urlbar/UrlbarInput.jsm
+++ b/mozilla-release/browser/components/urlbar/UrlbarInput.jsm
@@ -1092,16 +1092,14 @@ class UrlbarInput {
     this._resultForCurrentValue = null;
 
     // DB-2378 => CLIQZ-SPECIAL: do not set url if freshtabURL exists,
-    // do not check if custom URL is set by user
-    const homeUrl = Services.prefs.getCharPref("browser.startup.homepage", "");
-    if (
-      val.startsWith("moz-ext") &&
-      val === homeUrl
-    ) {
-      this.inputField.value = "";
-    } else {
-      this.inputField.value = val;
+    let setVal = val;
+    if (val.startsWith("moz-ext")) {
+      const homeUrl = Services.prefs.getCharPref("browser.startup.homepage", "");
+      if (val === homeUrl) {
+        setVal = "";
+      }
     }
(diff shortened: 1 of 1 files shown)


mozilla-release/browser/components/urlbar/UrlbarInput.jsm
```diff
@@ -1090,7 +1090,19 @@ class UrlbarInput {
 
     this.valueIsTyped = false;
     this._resultForCurrentValue = null;
-    this.inputField.value = val;
+
+    // DB-2378 => CLIQZ-SPECIAL: do not set url if freshtabURL exists,
+    // do not check if custom URL is set by user
+    const homeUrl = Services.prefs.getCharPref("browser.startup.homepage", "");
```
remusao: What about reading the pref only if `val.startsWith("moz-ext")` (which I assume will be false in more than 99% of cases). Then there would be no overhead at all?

alver-cliqz: then we will need to add one more condition

remusao: What about:
```js
    // DB-2378 => CLIQZ-SPECIAL: do not set url if freshtabURL exists,
    // do not check if custom URL is set by user
    if (
      val.startsWith("moz-ext") &&
      val === Services.prefs.getCharPref("browser.startup.homepage", "")
    ) {
      this.inputField.value = "";
    } else {
      this.inputField.value = val;
    }
```

alver-cliqz: I suggested:

alver-cliqz: this.inputField.value = val;
    if (val.startsWith("moz-ext")) {
      const homeUrl = Services.prefs.getCharPref("browser.startup.homepage", "");
      if (val === homeUrl) {
        this.inputField.value = "";
      }
    }

remusao: Up to you guys!

alver-cliqz: Good idea anyway, we will fix this

PR #1441, opened 2020-01-20
Restore backward compatibility of Linux builds with 16.04 and 18.04.

remusao:
The previous solution of using Ubuntu 19.10 for building was solving the
compilation issue with clang but introduced a backward incompatibility
because of the version of glibc available on the system.

The solution is to use the oldest version of Ubuntu that we want to
support (i.e. 16.04), then install custom tooling on top to allow
building. This is done by installing:

  * LLVM/Clang in their version 9 (latest)
  * Node.js LTS 12.x
  * Aptly from third-party ppa
  * Awscli from pypi using pip
  * Nasm from source
  * Rust toolchain using rustup

This solution gives us a frame to evolve the build of Cliqz for Linux in
the future. We can:

  * Bump the base image version to stop supporting an old version of Ubuntu
  * Bump versions of individual tools (i.e. Rust, Node.js, Clang, etc.)


philipp-classen:
Looks good at first glance. If it builds, it is safe to push to beta from my side.

PR #1445, opened 2020-01-23
linux: enable rust-simd compile flag

remusao:
This PR enables `rust-simd` compile flag which could bring some performance improvement. I already tested building the browser with this option and will now check if the resulting binary runs fine on different Linux distributions.


remusao:
@alver-cliqz I have updated the PR as requested by enabling the `--enable-rust-simd` compile flag for all platforms.

PR #1447, opened 2020-01-23
DB-2389: enable logging during runtime

aleksei-cliqz:


magic_build_and_package.sh
```diff
@@ -47,6 +52,19 @@ fi
 echo '***** Building *****'
 ./mach build
 
+if [ "$CQZ_INJECT_LOGGING" == "1" ]; then
+  echo '***** Injecting Logging *****'
+  cd ..
+  loggerConfigFile=cliqz-logger-config-mac.json
```
remusao: What about Linux?

aleksei-cliqz: > What about Linux?

I believe both MacOS and Linux belong to UNIX-like systems. So mac config will go well for Linux as well

cliqz-logger.js
```diff
@@ -0,0 +1,419 @@
+// First two arguments represent absolute path to node executable itself and
+// to cliqz-logger.js file.
+// So we are interested in arguments from the 3rd one;
+const inputArgs = Array.prototype.slice.call(process.argv, 2);
+
+const writeToConsole = function() {
```
remusao: What about:
```suggestion
const writeToConsole = function(...messages) {
```

cliqz-logger.js
```diff
@@ -0,0 +1,419 @@
...
+      configPath = inputArgs[i + 1];
+    } else {
+      // Places also can be augmented through the command line terminal;
+      places.push(inputArgs[i]);
+    }
+
+    i += 1;
+  }
+
+  new Promise(function(resolve, reject) {
```
remusao: Why create a `Promise` if nothing waits for it?

cliqz-logger.js
```diff
@@ -0,0 +1,419 @@
...
+          `"${data[i].lineNumber + i}", "${data[i].predicate}"` +
+        `);"${loggerVarName}";`;
+    }
+    source = source.join("");
+    // After injection is completed we write sources back to resolvedPath;
+    writeDataToPath(resolvedPath, source);
+  });
+};
+
+const readDataFromPath = function(path, cb) {
```
remusao: Why implementing it as a stream and not `fs.readFile` or `fs.readFileSync`? Eventually you will be joining all batches anyway.

cliqz-logger.js
aleksei-cliqz: @remusao In browser project we do not use latest ES6 features for now.
So I will stick to ES5 in this case.

remusao: Well, I can understand why not in browser code but this is a stand-alone script. Anyways, you guys decide :)

aleksei-cliqz: Starting from node.js v8.11.3 is should work fine.

cliqz-logger.js
aleksei-cliqz: Because it is easier to organize the code inside in that case since reading a file is asynchronous.

cliqz-logger.js
aleksei-cliqz: Because using stream always takes less memory to read the file.

remusao:
@aleksei-cliqz Would it be possible to see a diff of the line added after running this script on the code-base please?

aleksei-cliqz:
> @aleksei-cliqz Would it be possible to see a diff of the line added after running this script on the code-base please?

Actually, it would be. But for the browser we use a special trick to avoid diffs in the code.
We apply this script for the files belonging to a special directory ./obj
Git does not know anything about it. And in the browser this is what we need.
If you want to see diffs intentionally you need to run it explicitly in the terminal like this:

node ./cliqz-logger.js ./mozilla-release/browser/components/BrowserGlue.jsm
In that case the script would be executed only for the single file above which is not under ./obj directory. So you can see the diff.

mozilla-release/toolkit/modules/CliqzLogger.jsm
```diff
@@ -0,0 +1,48 @@
...
+          let messages = Array.prototype.slice.call(arguments);
+
+          messages.unshift(modulePath);
+          console.log.apply(console, messages);
+        }
+      };
+    },
+    destroy: function() {
+      Services.prefs.removeObserver(PREF_NAME, prefChangeObserver);
+      level = null;
```
remusao: Maybe this should be `0`?

magic_build_and_package.sh
remusao: Is it possible to maybe disable this mechanism completely for Linux?

mozilla-release/toolkit/modules/CliqzLogger.jsm
aleksei-cliqz: @remusao I am not sure that Garbage Collector will come after 0 values.

magic_build_and_package.sh
aleksei-cliqz: > Is it possible to maybe disable this mechanism completely for Linux?

Yes, we will disable that for Linux completely.

PR #1466, opened 2020-02-13
Fix Dockerfile for Linux build

remusao:
Build of Cliqz on Linux started failing a few days ago. It seems to be due to an incompatibility with `libc` that is shipped by default with Ubuntu 16.04 which we used until now to compile and source code from upstream Firefox/build toolchain. The reason we picked such an old version of Ubuntu to compile Cliqz at the time was the following:

* The `libc` version is bound to the linux distribution (and versions shipping from newer distributions usually introduce breaking changes).
* Version of `libc` will determine the compatibility of Cliqz for older systems (Cliqz compiled with `libc` from Ubuntu 18.04 will not be able to run on Ubuntu 16.04).

Hence, we picked the oldest version we could, so that Cliqz can run on as many distributions as possible. It would be possible to have custom building pipelines for each distributions but this is not a cost we can pay at the moment as it would introduce tremendous complexity compared to the "simple" setup we have now, thus we need to keep a single build for multi-linux support.

This PR fixes building on Linux, but will also force us to drop compatibility with some older systems (some further testing is required to have the final list but likely Ubuntu 16.04, some older Debian version, etc.). This compatibility could be re-introduced by using something by Flatpak/Snap though, which would likely bundle their own libraries (i.e. `libc`).

Cc @philipp-classen @alver-cliqz

### DIFF
--- a/cliqz_env.sh
+++ b/cliqz_env.sh
@@ -122,6 +122,9 @@ fi
 # set our own BUILD_ID in new build system, must be specified in format %Y%m%d%H%M%S
 export MOZ_BUILD_DATE=$CQZ_BUILD_ID
 
+export MOZ_SOURCE_REPO=https://github.com/cliqz-oss/browser-f
+export MOZ_SOURCE_CHANGESET=$CQZ_BUILD_ID
+
 # set path on S3 with BUILD_ID. From this path we take *.xpi and upload
 # build artifacts back (to locale folder, same as FF)
 export S3_UPLOAD_PATH=`echo dist/$CQZ_RELEASE_CHANNEL/$CQZ_VERSION/$MOZ_BUILD_DATE`
@@ -141,4 +144,3 @@ export CQZ_ADULT_DOMAINS_BF=../adult-domains.bin
 
 export ROOT_PATH=$PWD
 export SHELL=$SHELL
-
--- a/cliqz_env.sh
+++ b/cliqz_env.sh
@@ -122,9 +122,6 @@ fi
 # set our own BUILD_ID in new build system, must be specified in format %Y%m%d%H%M%S
 export MOZ_BUILD_DATE=$CQZ_BUILD_ID
 
-export MOZ_SOURCE_REPO=https://github.com/cliqz-oss/browser-f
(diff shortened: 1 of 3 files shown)


alver-cliqz:
Based on Mozilla's log for Linux build they are using different versions for:
* clang 9.0.1
* rust 1.39.0

cliqz_env.sh
```diff
@@ -122,6 +122,9 @@ fi
 # set our own BUILD_ID in new build system, must be specified in format %Y%m%d%H%M%S
 export MOZ_BUILD_DATE=$CQZ_BUILD_ID
 
+export MOZ_SOURCE_REPO=https://github.com/cliqz-oss/browser-f
+export MOZ_SOURCE_CHANGESET=$CQZ_BUILD_ID
```
remusao: @alver-cliqz let me know if you know a better way to add back this `MOZ_SOURCE_CHANGESET`. Do you remember having removed this `MOZ_SOURCE_REPO` variable as well at some point?

alver-cliqz: we never use it before, as I remember

alver-cliqz: Also, I don't want to use this flags on Win and Mac platform now (it could bring us some unwanted changes), so it they will work - let's use them only for linux build (not cross-build for Mac OS)

remusao: In any case, it does not seem to work. The only way I could make it "work" was by injecting values directly in `mozilla-release/toolkit/mozapps/installer/upload-files.mk` (line 184 and 185). I thought that the build system would pick up the env variables to populate this `source-repo.h` file (used from `upload-files.mk` but that does not seem to be the case).

alver-cliqz: @chrmod , can you please have a look. As I see - nothing changes in build scripts after merge with Firefox 73. Mai changes only switch from Python 2 to Python 3.

remusao: The more likely explanation is that this is not due to any Firefox upstream change but the bump from Ubuntu 16.04 to Ubuntu 18.04. It could be that these two fields were optional when creating the RPM package but with a newer version they became mandatory.

PR #1476, opened 2020-02-20
DB-2419: remove normandy from build

tsl143:


alver-cliqz:
I'm not sure this is proper way to turn it off. Let's check which functionality is involved into build process with this variable, I think at least updating extension's blocklist can be affected (for now we still using Mozilla's databases and update process)

tsl143:
based on Alex's comment, closing this one

PR #1482, opened 2020-03-20
DB-2431: disable geolocation prefs | all.js

aleksei-cliqz:


mozilla-release/modules/libpref/init/all.js
```diff
@@ -4047,16 +4048,16 @@ pref("network.psl.onUpdate_notify", false);
 pref("geo.provider.network.timeout", 60000);
 
 #ifdef XP_MACOSX
-  pref("geo.provider.use_corelocation", true);
+  pref("geo.provider.use_corelocation", false);
```
tsl143: @aleksei-cliqz we had this true in `1.33`, are you sure to make this false?

aleksei-cliqz: > @aleksei-cliqz we had this true in `1.33`, are you sure to make this false?

That is what we had in 1.33.1
`#ifdef XP_MACOSX
  pref("geo.provider.use_corelocation", false);
#endif

// Set to false if things are really broken.
#ifdef XP_WIN
  pref("geo.provider.ms-windows-location", false);
#endif

#if defined(MOZ_WIDGET_GTK) && defined(MOZ_GPSD)
  pref("geo.provider.use_gpsd", false);
#endif`

aleksei-cliqz: df42477b28a72116811c112b684966b6e4b707ce

tsl143: My bad, I got confused with something else, apologies.

aleksei-cliqz: > My bad, I got confused with something else, apologies.

No worries, that is why we work together as a team man;)

mozilla-release/modules/libpref/init/all.js
```diff
@@ -4034,6 +4034,7 @@ pref("network.psl.onUpdate_notify", false);
   pref("widget.wayland_vsync.enabled", false);
 #endif
 
+// CLIQZ-SPECIAL - force mozilla location service only
 // All the Geolocation preferences are here.
 //
 #ifndef EARLY_BETA_OR_EARLIER
```
alver-cliqz: Please, use same code as we had in firefox.js here

PR #1483, opened 2020-03-20
DB-2416: fixing AFW behavior

tsl143:


mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -723,6 +723,25 @@ HttpObserverManager = {
         this.handleForgetModeNotification(otherWin);
       }
 
+      try {
```
aleksei-cliqz: @tsl143 Please add our comments tag #CLIQZ-SPECIAL
Also I thought that it might be a good idea to not only add the tag itself but with the browser version at the end as well.
#CLIQZ-SPECIAL-1.34

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -737,8 +756,21 @@ HttpObserverManager = {
 
       if (isSecondaryTab) {
         gBrowser.removeTab(isSecondaryTab);
+      } else {
+        try {
```
aleksei-cliqz: @tsl143 Please add our comment tag with the browser version.

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
aleksei-cliqz: Also could you please tell which part might throw an error here? Why do you need try/catch?

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
aleksei-cliqz: And also the same question about try/catch. Do you really need this?

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -723,6 +723,25 @@ HttpObserverManager = {
...
+        // Covers the case of balnk window in case explicit link is opened in new tab from context menu
+        const { BrowserWindowTracker } = ChromeUtils.import("resource:///modules/BrowserWindowTracker.jsm");
+        BrowserWindowTracker.orderedWindows.forEach(w => {
+          const { tabs } = w.gBrowser;
+          if (
+              (
+                tabs[0]._fullLabel === originURL || // for cases like google
+                !tabs[0]._fullLabel // for cases like duckduckgo
+              ) &&
+              tabs.length === 1
```
aleksei-cliqz: @tsl143 also could you please tell what happens if a browser has more than 1 tab?
For example, porno-movie.domain will still be opened in the original tab and also in a forget window?

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: For try catch, I am expecting always the `tab[0]` exist, in weirdest case or by any future release this may change.

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: same as below comment

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: In this case there must not be second tab, as we will land in this situation only when user opens it via `open in new window`

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: as above

aleksei-cliqz: > For try catch, I am expecting always the `tab[0]` exist, in weirdest case or by any future release this may change.

Then why not to check just its' existence? 
if (tabs[0] != null) {
  // here goes your code if tabs[0] exists;
}

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: for version tag, lets discuss this offline as this is beyond scope of this PR

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: yeah, true, but I am just following try catch scheme here for different checks, so sticking to it.

aleksei-cliqz: Ok. Then how can we know about the error? It will be silently ignored and swallowed by catch.
Either throw it explicitly in catch or use that tabs[0] against null check please.
I really do not want to fall into situation similar to what we had in SessionStore again.

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
aleksei-cliqz: > for version tag, lets discuss this offline as this is beyond scope of this PR

Sure.

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -723,6 +723,25 @@ HttpObserverManager = {
...
+            ) {
+            w.gBrowser.removeTab(tabs[0]);
+          }
+        });
+      } catch(e){
```
aleksei-cliqz:
```suggestion
      } catch(e){
         console.error('maybeAFW: error thrown | ', e);
```

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -737,8 +756,21 @@ HttpObserverManager = {
...
+      } else {
+        try {
+          // Deletes the tab if there is no history path
+          if (!gBrowser.webNavigation.canGoBack) {
+            if (tabs.length === 1) {
+              openTrustedLinkIn(freshTabURL, "tab");
+            }
+            gBrowser.removeTab(selectedTab);
+          }
+        } catch(e){
```
aleksei-cliqz:
```suggestion
        } catch(e){
          console.error('maybeAFW: error thrown | ', e);
```

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
tsl143: We have added console to catch block, for better error deduction.

mozilla-release/toolkit/components/extensions/webrequest/WebRequest.jsm
```diff
@@ -723,6 +723,26 @@ HttpObserverManager = {
         this.handleForgetModeNotification(otherWin);
       }
 
+      try {
+        // Covers the case of balnk window in case explicit link is opened in new tab from context menu
```
tsl143:
```suggestion
        // Covers the case of blank window in case explicit link is opened in new window from context menu
```

PR #1484, opened 2020-03-20
DB-2432: turn off interventions provider by default

aleksei-cliqz:


tsl143:
Do you think it will be good idea to remove `UrlbarProviderInterventions` from `moz.build` and `jar.mn` ?

aleksei-cliqz:
> Do you think it will be good idea to remove `UrlbarProviderInterventions` from `moz.build` and `jar.mn` ?

I do not think so for now. Because there are many other providers along the way which also have to be disabled following this case.
It is easier to control it via prefs for now.

mozilla-release/browser/app/profile/firefox.js
```diff
@@ -371,7 +371,10 @@ pref("browser.urlbar.usepreloadedtopurls.expire_days", 14);
...
   // If true, we show actionable tips in the Urlbar when the user is searching
   // for those actions.
-  pref("browser.urlbar.update1.interventions", true);
+
+  // CLIQZ-SPECIAL: DB-2432, in Cliqz Browser this always has to be false
+  // since we use our own extension dropdown.
+  pref("browser.urlbar.update1.interventions", false);
   // If true, we show new users and those about to start an organic search a tip
   // encouraging them to use the Urlbar.
   pref("browser.urlbar.update1.searchTips", true);
```
alver-cliqz: May be turn off this also?

alver-cliqz: Ah, no needs. It disabled for both our builds, beta and release. Please create a ticket for next version (1.35) to check this

alver-cliqz:
We don't need to do anything with it now, because it disabled for both our builds, beta and release. Please create a ticket for next version (1.35) to check this

PR #1490, opened 2020-03-23
DB-2426: fix signing process for Mac build

alver-cliqz:


Jenkinsfile.macoslinux.crossbuild
```diff
@@ -189,6 +191,8 @@ image.inside("-w /opt/browser-f"){
     stage('Unstash signed artifacts'){
         unstash 'signeddedmg'
```
chrmod: it does  not seem we still use DMGs, this can be removed to speed up the build

sign_mac.sh
```diff
@@ -6,28 +6,49 @@ set -x
...
+  xattr -cr $PKG_DIR/Cliqz.app
+  xattr -cr $PKG_DIR/Cliqz.app/Contents/MacOS/plugin-container.app
+  xattr -cr $PKG_DIR/Cliqz.app/Contents/MacOS/crashreporter.app
+  xattr -cr $PKG_DIR/Cliqz.app/Contents/MacOS/updater.app
+
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/plugin-container.app
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/pingsender
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/libfreebl3.dylib
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/cliqz-bin
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/liblgpllibs.dylib
```
chrmod: Could we codesign all `dylibs` with a glob pattern, wildcard or for loop?

sign_mac.sh
```diff
@@ -6,28 +6,49 @@ set -x
...
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/libnss3.dylib
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/minidump-analyzer
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/crashreporter.app
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/crashreporter.app/Contents/MacOS/crashreporter
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/updater.app
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/MacOS/updater.app/Contents/MacOS/org.mozilla.updater
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/Library/LaunchServices/org.mozilla.updater
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app/Contents/Resources/gmp-clearkey/0.1/libclearkey.dylib
+  codesign "${SIGN_ARGS[@]}" $PKG_DIR/Cliqz.app
+  codesign -vvv --deep --strict $PKG_DIR/Cliqz.app
```
chrmod: Isn't `--deep` already codesigning everything inside?

sign_mac.sh
alver-cliqz: I'm not good in bash scripting

sign_mac.sh
alver-cliqz: Unfortunately, not. Also please check args which we use for signing. I just repeat all steps which Mozilla's do for Firefox. All this step is necessary for proper notarization afterward

Jenkinsfile.macoslinux.crossbuild
alver-cliqz: Yes, this is my thoughts also, I will switch to use only apps

alver-cliqz: Umm. I tried to play with this idea - it's hard, because processing DMG is not native on Linux platform, so for packing-unpacking I need to investigate DMG_TOOL, MKFSHFS and HFS_TOOL tools. It absolutely worth to learn them and switch to using it, but now I need to make quick solution for new version

Jenkinsfile.macoslinux.crossbuild
```diff
@@ -189,6 +212,17 @@ image.inside("-w /opt/browser-f"){
...
+        unstash 'signeddeapp'
+        unstash 'signedusapp'
+         sh """#!/bin/bash -l
+            set -e
+            set -x
+            cd obj/dist/cliqz
+            tar -jxvf cliqz.app.en-US.tar.bz2
+            cd ../l10n-stage/cliqz
+            tar -jxvf cliqz.app.de.tar.bz2
+            cd ../../../../
```
chrmod: You don't have to do that as this shell is going to be kill anyways

PR #1511, opened 2020-04-03
DB-2447_2: SessionStore remove duplicated freshtabs | fix

aleksei-cliqz:


mozilla-release/browser/components/sessionstore/SessionStore.jsm
```diff
@@ -488,6 +488,9 @@ var SessionStore = {
 Object.freeze(SessionStore);
 
 const cliqz_removeDuplicatedEntries = function(tabs, overwriteTabs) {
+  if (!Array.isArray(tabs)) {
```
tsl143, line 491: what are expected value of tabs? 
`undefined` can be dealt with default value `tabs = []`;

aleksei-cliqz: @tsl143 We always expect tabs to be an array.
But expectation can not guarantee it will be that.
Let's leave this check instead of default value in args.
Since it is more obvious.

tsl143: Okay, then let's use `return true`

aleksei-cliqz: I am not sure we can return true here since winData.tabs expects array.
Please take a look down below at the place this function is called.

PR #1518, opened 2020-04-08
DB-2444: enable dnt by default

tsl143:


tsl143:
This is still under discussion

PR #1567, opened 2020-07-08
DB-2513: merge with Firefox 78.0.1

alver-cliqz:


mozilla-release/browser/branding/cliqz/branding.nsi
```diff
@@ -20,6 +20,9 @@
 ; Beta since they share the same branding when building with other branches that
 ; set the update channel to beta.
 !define OFFICIAL
+!define URLStubDownloadX86 "https://download.mozilla.org/?os=win&lang=${AB_CD}&product=firefox-latest"
```
chrmod, line 23: Not sure if we need this, even if - it should probably not mention mozilla.org

alver-cliqz: We don't use this code now. This is part of mini-installer from Firefox. They re-worked it a lot, and now it could be count as totally different NSIS script. Unfortunately, our old mini-installer not possible to build (wierd problem with NSIS configs or so). I don't want to waste time on it right now, because we do not planning to distribute new mini-installer anywhere.

